PR #2, opened 2020-02-03
Add zones and options to backup and restore volume info

disrani-px:


prashanthpx:
LGTM. We also need to remove references to initial state in the controller code.

PR #8, opened 2020-02-18
Added status parameter to the cluster object.

siva-portworx:
- Added status parameter to the cluster object to store the status of cluster. Possible values are online and offline.


pkg/apis/v1/api.proto
```diff
@@ -46,6 +46,17 @@ message ClusterInfo {
...
+    StatusInfo status = 4;
+
+    // Message for maintaing status of the cluster.
+    message StatusInfo {
+        enum Status {
+            Invalid = 0;
+            Online = 1;
+            Offline = 2;
+        }
+        Status status = 1;
```
disrani-px: We'll probably need a string Reason field here to display info about why it might be offline

siva-portworx: fixed.

PR #13, opened 2020-03-11
Adding status field for backuplocation.

siva-portworx:
Adding status field for backuplocation to store Deleting or Delete Pending status.


pkg/apis/v1/api.proto
```diff
@@ -258,7 +258,21 @@ message BackupLocationInfo {
         (gogoproto.moretags) = "secure:\"true\""
     ];
     string cloud_credential = 4;
+    StatusInfo status = 5;
+    bool cloud_delete = 6;
```
prashanthpx: Should you call these as `backup_delete`? If a user is hosting something like minio, the term cloud_delete could be confusing.

prashanthpx: Or even like `delete_backups`

siva-portworx: Made it as delete_backups.

pkg/apis/v1/api.proto
```diff
@@ -258,7 +258,21 @@ message BackupLocationInfo {
...
+    StatusInfo status = 5;
+    bool delete_backups = 6;
     oneof config { S3Config s3_config = 100; }
+    // Message for maintaing status of the object.
+    message StatusInfo {
+        enum Status {
+            Invalid = 0;
+            Valid = 1;
+            Deleting = 2;
+            DeletePending = 3;
```
disrani-px:
```suggestion
            Deleting = 2;
            DeletePending = 3;
```

Not clear why we need both. Add comments or use just one status

siva-portworx: As discussed, keeping only DeletePending state alone

PR #22, opened 2020-03-22
Added backupBy variable tobe part of backup object.

siva-portworx:
Added backupBy variable tobe part of backup object.

        - BackupBy variable will have values as px-backup or stork.
        - It is used to determine whether the backup was taken through px-backup or
          through stork directly.
        - It will be used to determine backup CR name format while deleting the
          backup CR.


pkg/apis/v1/api.proto
```diff
@@ -280,6 +280,13 @@ message BackupInfo {
     // px-backup.
     // Otherwise it will be empty.
     BackupSchedule backup_schedule = 12;
+    BackupBy backup_by = 13;
```
disrani-px: CreatedBy would be more apt

siva-portworx: Changed it to CreatedBy

PR #31, opened 2020-04-14
Add azure backup location config

stgleb:


stgleb:
@disrani-px could you elaborate on that? I see that right now px-backup does not allow me to create azure backup location.

disrani-px:
You need to create a cloud credential for Azure and then point to that cloud credential from the BackupLocation

stgleb:
Okay, I'll try it

PR #33, opened 2020-04-25
Add kube auth plugins to fetch credentials from grpc server

disrani-px:


pkg/kubeauth/gcp/gcp.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+type gcpToken struct {
+	tokenSource oauth2.TokenSource
+}
+
+// Init initializes the gcp auth plugin
+func (g *gcp) Init(conn *grpc.ClientConn) error {
+	g.conn = conn
+	if err := rest.RegisterAuthProviderPlugin(pluginName, g.newGCPAuthProvider); err != nil {
```
ram-infrac, line 40: let's make sure we register auth provider once. Since if client called init for multiple time client-go rest through error ```Auth Provider Plugin \"gcp\" was registered twice```

or let's move grpc connection from  Init to UpdateClient

disrani-px: Removed `Init` and added `conn` to `UpdateClient()`

PR #34, opened 2020-04-26
PB-414: Added license consumed count

prashanthpx:
As part of InspectResponse added number of licenses consumed.


pkg/apis/v1/api.proto
```diff
@@ -1343,7 +1343,10 @@ message LicenseActivateResponse {}
 message LicenseInspectRequest { string org_id = 1; }
 
 message LicenseInspectResponse {
-    repeated LicenseResponseInfo response_info = 1;
+    // Number of license consumed
+    int64 consumed = 1;
```
disrani-px, line 1347: This needs to be per feature

prashanthpx: As discussed made the changed.

pkg/apis/v1/api.proto
```diff
@@ -1342,19 +1342,28 @@ message LicenseActivateResponse {}
...
+        repeated EntitlementInfo entitlement_info = 3;
+    }
+
+    message EntitlementInfo {
+        // License count
+        int64 count = 1;
+        // Expiry time of license
+        google.protobuf.Timestamp expires = 3;
+        // Start time of license
+        google.protobuf.Timestamp starts = 4;
```
disrani-px, line 1365:
```suggestion
        google.protobuf.Timestamp expires = 2;
        // Start time of license
        google.protobuf.Timestamp starts = 3;
```

PR #42, opened 2020-06-15
PB-564: Added size field to BackupInfo message

prashanthpx:
Added backup_size and size fields to backup object.


pkg/apis/v1/api.proto
```diff
@@ -299,6 +299,7 @@ message BackupInfo {
     // Otherwise it will be empty.
     BackupSchedule backup_schedule = 12;
     string cr_name = 13;
+    uint64 backup_size = 14;
```
disrani-px, line 302:
```suggestion
    uint64 size = 14;
```

prashanthpx: Fixed.

PR #59, opened 2020-08-27
Move incremental count to a oneof type

disrani-px:


pkg/apis/v1/api.proto
```diff
@@ -148,7 +147,7 @@ message SchedulePolicyInfo {
         int64 retain = 2;
         // Number of incremental snapshots to take before taking a full
         // snapshot.
-        google.protobuf.UInt32Value incremental_count = 3;
+        oneof incremental_count_set { uint64 incremental_count = 4; }
```
prashanthpx, line 150: Should be 3.

PR #65, opened 2020-09-23
Add API support for BackupDeleteRequest target cluster ID

ggriffiths:
Signed-off-by: Grant Griffiths <grant@portworx.com>


ggriffiths:
To be used here for choosing a cluster to delete through:
https://github.com/portworx/px-backup/blob/master/pkg/server/backup.go#L440

disrani-px (approved):

PR #66, opened 2020-10-05
Adding support for payg

prashanthpx:


pkg/apis/v1/api.proto
```diff
@@ -1453,6 +1464,16 @@ message LicenseResponseInfo {
         google.protobuf.Timestamp expires = 3;
         // Start time of license
         google.protobuf.Timestamp starts = 4;
+        // License type
+        // Either "Usage Based" or normal license
+        string type = 5;
```
disrani-px, line 1469: Use enum instead of string type

prashanthpx: Fixed.

disrani-px (approved):

PR #68, opened 2020-11-06
fixed the compilation error due to grpc-ecosystem pkg.

siva-portworx:
- Added change to use latest v1 version (1.16.0) of grpc-ecosystem pkg.
- With GOPATH, we can not do go get of particular version, So added changes to use go module.


disrani-px (commented):
Gopkg.lock and Gopkg.toml should be removed

siva-portworx:
> Gopkg.lock and Gopkg.toml should be removed

Deleted the Gopkg.lock and Gopkg.toml and updated the PR.

disrani-px (approved):

PR #69, opened 2020-11-09
Added changes to include delete backuplocation flag for cloud cred deletion request.

siva-portworx:
Added changes to include delete backuplocation flag for cloud cred deletion request.
Also added cloud credit name to filter backup location enumerate based on the cloud cred.


siva-portworx:
With the latest flow, we do not need deleteBackuplocation flag in cloud credential delete API. So closing this PR.

PR #86, opened 2021-03-11
Adding support for passing resource types in backup

prashanthpx:
Support to pass resource types as part of backup and schedules.

[Note: Due to formatting (which is not right on master), the diff looks huge.  Changes are mainly in bakcupcreate and schedulecreate requests]


pkg/apis/v1/api.proto
```diff
@@ -1089,14 +1159,16 @@ message BackupCreateRequest {
     repeated ResourceInfo include_resources = 8;
     // Snapshot class name to use for backup (optional)
     string csi_snapshot_class_name = 9;
+    // List of resource types to backup (optional)
+    repeated string resource_types = 10;
```
siva-portworx, line 1163: Are we are not storing the resourceTypes from the request in to backup/backupschedule objects?
It will be help for debug. we do that for include_rescources.

prashanthpx: Yes we can add.

siva-portworx (approved):

PR #91, opened 2021-04-09
Add gRPC service & proto definitions for Role Object

ram-infrac:
This PR adds -
 - proto struct definitions for role object
 - gRPC service definitions for role

Closes #


pkg/apis/v1/api.proto
```diff
@@ -1589,4 +1589,99 @@ message ActivityEnumerateResponse {
...
+    // Metadata of the object
+    Metadata metadata = 1 [ (gogoproto.embed) = true ];
+    // ServiceRule associated with role object
+    repeated ServiceRule rules = 2;
+}
+
+// ServiceRule object for set of rules associated with each role object
+// services - PX-Backup services names
+// apis - allowed apis for px-backup services
+message ServiceRule {
```
ram-infrac, line 1604: Since `Rule` is already used for backup pre/post exec rules using `ServiceRule` for Role Objects rule,  kindly suggest better name suggestions if any

pkg/apis/v1/api.proto
```diff
@@ -1589,4 +1589,99 @@ message ActivityEnumerateResponse {
...
+    // The gRPC service name
+    repeated string services = 1;
+    // The API name in the service in lowercase
+    repeated string apis = 2;
```
siva-portworx, line 1608: In the custom role definition, do allow to configure specific combination of API for a specific service. I mean 
backuplocation with View only but schedulepolict with all option?

ram-infrac: yes you will need to define role as  -
```
Role {
 {
    "services": {"backup location"},
    "api" : {"inspect*", "enumerate*"},
}
{  "services: { "SchedulePolicy"},
    "api:" {"*"},
    }
}
```
^^  REST equivalent

pkg/apis/v1/api.proto
```diff
@@ -1589,4 +1589,99 @@ message ActivityEnumerateResponse {
...
+    rpc Delete(RoleDeleteRequest) returns (RoleDeleteResponse) {
+        option (google.api.http) = {
+            delete : "/v1/role/{org_id}/{name}"
+        };
+    }
+}
+
+// RoleCreateRequest defines roleobject create structure
+// Metadata with each role object
+// ServicRule set of rules over api services for given role object
```
siva-portworx, line 1649: typo: ServicRule

pkg/apis/v1/api.proto
```diff
@@ -1589,4 +1589,99 @@ message ActivityEnumerateResponse {
...
+message RoleCreateRequest {
+    CreateMetadata metadata = 1 [ (gogoproto.embed) = true ];
+    repeated ServiceRule rules = 2;
+}
+
+message RoleCreateResponse {}
+// RoleUpdateRequest efines roleobject update structure
```
siva-portworx, line 1656: typo: efines

pkg/apis/v1/api.proto
siva-portworx: RoleRule :)

siva-portworx (approved):

pkg/apis/v1/api.proto
siva-portworx: ok got it. Thanks

adityadani (approved):
looks good once the typos are fixed

pkg/apis/v1/api.proto
ram-infrac: I think we should go with `RoleConfig`

PR #92, opened 2021-04-11
PB-1244: Adding ownership object replacing owner in Metadata structure

prashanthpx:
Adds ownership object to Metadata section relacing the exiting Ownership (a name)

PB-1244

This PR would not be pushed to master till other changes in px-backup are done, doing so will break px-backup when someone vendors top of master

### DIFF
--- a/pkg/apis/v1/common.proto
+++ b/pkg/apis/v1/common.proto
@@ -28,6 +28,10 @@ message Metadata {
     string name = 1;
     // uid for the object
     string uid = 2;
+    // owner of the object. This field is deprecated, not to be used
+    // TODO: Currently gRPC supports marking field as [deprecated=true]
+    // As we need to access this field as part of migration to new Ownership
+    // message will not use the same for now.
     string owner = 3;
     // organization uid
     string org_id = 4;
@@ -37,6 +41,8 @@ message Metadata {
     map<string, string> labels = 7;
     // create time in sec
     int64 create_time_in_sec = 8;
+    // ownership of the object
+    Ownership ownership = 9;
 }
 
 message CreateMetadata {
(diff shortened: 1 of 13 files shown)


pkg/apis/v1/common.proto
```diff
@@ -45,7 +45,65 @@ message CreateMetadata {
     // org id of the object
     string org_id = 2;
     // owner of the object
-    string owner = 3;
+    Ownership ownership = 3;
```
siva-portworx, line 48: Don't we need to have ownership field in metadata definition as well? So that it will be stored with each instance of object.

siva-portworx (approved):

pkg/apis/v1/common.proto
prashanthpx: Yes needed. Added it.

pkg/apis/v1/common.proto
```diff
@@ -45,7 +46,65 @@ message CreateMetadata {
...
```
siva-portworx, line 49: Will it not affect serialization, if we change the existing field to of different type? I mean when we try read the old object with this new definition.

pkg/apis/v1/common.proto
```diff
@@ -28,7 +28,8 @@ message Metadata {
     string name = 1;
     // uid for the object
     string uid = 2;
-    string owner = 3;
```
adityadani, line 31: This will break backward compatibility. In general, when using protobufs one should not replace a field type/number.
We will have to keep the owner field and mark it as Deprecated and introduce a new Ownership field.

pkg/apis/v1/common.proto
```diff
@@ -45,7 +46,65 @@ message CreateMetadata {
     // org id of the object
     string org_id = 2;
     // owner of the object
-    string owner = 3;
+    Ownership ownership = 3;
     // labels associated with the object
     map<string, string> labels = 4;
 }
+
+// Ownership information for objects(eg: bakcup object, schedule object).
```
adityadani, line 54: nit: backup

pkg/apis/v1/common.proto
```diff
@@ -45,7 +46,65 @@ message CreateMetadata {
...
+        map<string, AccessType> collaborators = 2;
+        // Public access to objects may be assigned for access by the public
+        // userd
+        // TODO: Instead of PublicAccessControl, can we just set '*' in
+        // collaborators with required permision?
+        PublicAccessControl public = 3;
+    }
+
+    // Username of owner
+    // The storage system uses the username taken from the security
```
adityadani, line 103: nit: Backup uses the username taken taken from the token...

adityadani (changes requested):

pkg/apis/v1/common.proto
prashanthpx: This is something which we need to see how to handle in migrating existing objects to new ones which need discussion.   
Removing a field in higher proto is an issue. 
One option I was thinking of is having the current `owner` as is but add another `OwnerShip` message separately and not use the current one.  As we discuss, we might come up with different options.

pkg/apis/v1/common.proto
prashanthpx: Yes, this is the thinking I had which I wanted to discuss hence i didn't pus the PR to master.

pkg/apis/v1/common.proto
siva-portworx: Actually remove a field is not an issue, unless, we try to use the same index in future. If we do not reuse remove field number, it should not cause any issue.

prashanthpx: Yes, we need to use different indexing.

adityadani (approved):

ram-infrac (approved):

PR #93, opened 2021-04-13
PB-1268: Add a new service for Managed Clusters.

adityadani:
-  Managed clusters are kubernetes clusters managed by providers like AWS, GCP,
AKS etc.
- The ManagedClusters service will provide APIs to interact with these
providers and their managed clusters.
- Currently the only supported API is Enumerate which will enumerate all the
clusters from the provider for a given cloud credential. The credential enforces
which clusters will be returned in the API based on its permissions.
- Since this service and its APIs act as a front for such providers they do not
have the Metadata object in their structs. This API will not return any objects
from the datastore but instead will return objects from runtime.

Closes #


pkg/apis/v1/api.proto
```diff
@@ -1542,6 +1542,51 @@ service License {
...
+    // Enumerate will return list of managed clusters that
+    // this cloud credential has access to
+    string cloud_credential = 3;
+}
+
+// Define ManagedClusterObject struct
+message ManagedClusterObject {
+    // Name of the managed cluster
+    // This is usually the user friendly name of the cluster
+    string name = 1;
```
adityadani, line 1572: The name and uid are the cloud's name and uid for the cluster. The caller of this API will use this information to send a Cluster.Create request.

pkg/apis/v1/api.proto
```diff
@@ -1542,6 +1542,51 @@ service License {
...
+    // UID of the managed cluster
+    // This is the UID returned by the managed cluster provider
+    string uid = 2;
+    // Cloud Credential name used for fetching this cluster's details
+    string cloud_credential = 3;
+    // Kubeconfig if the provider supports using kubeconfigs
+    string kubeconfig = 4 [
```
adityadani, line 1579: Certain cloud providers do not provide any API to fetch kubeconfigs, in such cases this field will be empty. The k8s rest.Config object needs to be refreshed at runtime periodically using the cloud credentials in such cases.

pkg/apis/v1/api.proto
```diff
@@ -1542,6 +1542,51 @@ service License {
     }
 }
 
+service ManagedClusters {
+
+    // Enumerate returns a list of managed clusters from a provider like
+    // AWS/other clouds
+    // using the given credentials
+    rpc Enumerate(ManagedClusterEnumerateRequest)
```
siva-portworx, line 1550: The managedCluster Enumerate will be used by GUI to get the list of cluster for the given cred and once the user selected a set of cluster from the list, the GUI will call normal cluster create for each of the cluster. Is that understanding is correct?

adityadani: Yes correct

siva-portworx: ok got it. Thanks. I was thinking how the Ux will look like to communicate the status of the actual cluster object addition, if the user selected more number of cluster name from the list and couple of the addition fails.

siva-portworx (approved):
LGTM

ram-infrac (approved):

pkg/apis/v1/api.proto
```diff
@@ -1542,6 +1542,51 @@ service License {
...
+    string org_id = 1;
+    bool include_secrets = 2;
+    // cloud credential name.
+    // Enumerate will return list of managed clusters that
+    // this cloud credential has access to
+    string cloud_credential = 3;
+}
+
+// Define ManagedClusterObject struct
+message ManagedClusterObject {
```
prashanthpx, line 1569: What is the use case to differentiate ManagedCluster and normal clusters? Was thinking if we could add a subfield in cluster object which says it's managed or not that way we had one ClusterObject for all clusters and the current framework takes care of ClusterObject life cycle management.

prashanthpx (approved):
LGTM

pkg/apis/v1/api.proto
siva-portworx: As such managedCluster is not persisted to datstore like normal cluster object. ManagedCluster Enumerate will just query and get the list of clusters from the providers and return to the caller. Px-backup does not store any information about the managed cluster. It is just a wrapper to the UI to call this Enumerate to show up the list of cluster that belongs to this cred in the given provider. Once the user selects the cluster, the UI will call normal cluster create.

adityadani: Correct. I wanted to also differentiate between objects that are stored in our datastore which require all CRUD operations and this object where we don't want all the CRUD operations.

pkg/apis/v1/api.proto
adityadani: Yes we will have to carefully handle that in the UI. From backup's point of view we should still keep a single cluster Add API. The UI can allow multi-select but should show clear errors which clusters failed. Wdyt ?

PR #98, opened 2021-04-20
PB-1227: Add IBM kubeauth support

adityadani:
The PR contains the following high level changes
**Unify kubeauth**
- Maintain a single kubeauth pkg that Backup and other middleware layers can use.
- Maintain two APIs in kubeauth
   - UpdateClient -> takes in a grpc client and fetches the credential object
   - UpdateClientByCredObject -> uses the provided credential object to get cloud credentials.

**Add support for IBM**
- Add IBM pkg to kubeauth
- Add IBM cloud type to Credential object
   - IBM Credentials takes in 1) API Key and 2) Region fields.
   - Since we only need to get the cluster info, we don't need IBM ResourceGroup
- Add logic to use IBM APIs to fetch a new kubeconfig and update the rest Config object.

Closes #


pkg/kubeauth/ibm/ibm.go
```diff
@@ -0,0 +1,328 @@
...
+		logrus.Debugf("failed to create config path: %v", err)
+	}
+	defer func() {
+		i.cleanupKubeconfigPath(configPath)
+	}()
+
+	target := v1.ClusterTargetHeader{
+		Region: region,
+	}
+	kubeConfigPath, err := clusterAPI.GetClusterConfig(clusterName, configPath, true, target)
```
adityadani, line 173: The 3rd argument if set to true, pulls the admin kubeconfig. @sharma-tapas What do you suggest we set here?

prashanthpx (approved):
The flow looks good to me.

pkg/apis/v1/api.proto
```diff
@@ -66,6 +66,13 @@ message AWSConfig {
     ];
 }
 
+message IBMConfig {
+    string api_key = 1 [
+        (gogoproto.jsontag) = "api_key",
+        (gogoproto.moretags) = "secure:\"true\""
+    ];
```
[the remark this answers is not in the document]
adityadani: For getting the cluster info region is not needed since we are parsing this cluster info - name and region from the uploaded kubeconfig itself.

pkg/kubeauth/ibm/ibm.go
sharma-tapas: Instead of `GetClusterConfig` use `GetClusterConfigDetail` and to get the path use `clusterInfo.FilePath`
For future, when we will manage the kubeconfigs
```If we know the cluster is an OCP cluster, do not pull the admin kube config, this ways it triggers the IBM SDK to fetch proper IAM token in the latest 4.6.23 release of OCP.```

File: pkg/kubeauth/ibm/ibm.go [hunk 3]
```diff
@@ -0,0 +1,327 @@
...
+	pluginName            = "ibm"
+	idpIssuerUrlKey       = "idp-issuer-url"
+	ibmIssuerUrlSubstring = "iam.cloud.ibm.com"
+)
+
+type ibm struct {
+	clusterMapLock map[string]*sync.Mutex
+}
+
+// Init initializes the gcp auth plugin
```
sharma-tapas, line 41: nit: Initialize the IBM auth plugin

sharma-tapas (approved):

PR #100, opened 2021-04-22
PB-1342: store keyclock role id with px-backup role info

ram-infrac:
- PB 1342: since keyclock only allow role uid for role mapping between user and role we will need to store `role id`
with px-backup role info
https://www.keycloak.org/docs-api/5.0/rest-api/index.html#_role_mapper_resource
- PB-1341: allow enumerate filters for role objects

Closes #


siva-portworx (approved):
As discussed, we will go ahead with this approach for role, when my PR for UID, i will fix to use the different structure that will have both uid and name.

PR #101, opened 2021-05-11
PB-1398: Add Role permission support 

ram-infrac:
Permission API support in role service

Closes # PB-1398


adityadani (approved):

pkg/apis/v1/api.proto
```diff
@@ -1771,3 +1777,9 @@ message RoleDeleteRequest {
 
 // Define RoleDeleteResponse struct
 message RoleDeleteResponse {}
+
+// Define RolePermissionRequest struct
+message RolePermissionRequest { string org_id = 1; }
+
+// Define RolePermissionResponse struct
+message RolePermissionResponse { map<string, bool> rules = 1; }
```
adityadani, line 1785: On a second thought returning a boolean is not extensible. What if in future we want to differentiate between Read and Write permissions? Can we instead change this to 
```suggestion
message Permission {
     FullAccess bool
     ViewAccess bool
}
message RolePermissionResponse { map<string, Permission> rules = 1; }
```
In this way later on we can split Full into even more sub permissions if we want to.

adityadani (changes requested):

pkg/apis/v1/api.proto
ram-infrac: user can either have full access/view access,
if we want to detailed out RolePermissionResponse, we should send response for each create/update/delete/list operation

adityadani: My point is if you have a `map<string, bool>` then it becomes hard to extend it.
Instead have a `map<string, struct{}>` so that the struct is extendable. Right now the struct can have a single boolean value like you have right now.

ram-infrac: make sense, will update it to `struct` thanks

adityadani (approved):

PR #111, opened 2021-08-19
pb-1844: Added following required proto changes for data mover feature.

siva-portworx:
```
 pb-1844: Added following required proto changes for data mover feature.

      1) Added backupType in BackupCreateRequest message definition.
      2) Added backupType in BackupScheduleCreateRequest message definition.
      3) Added backupType in BackupInfo and BackupScheduleInfo message
         definition.
      4) Added storage_class in the volume message definition in the
         backupInfo message definition.
      5) Added storage_class_mapping map in RestoreCreateRequest message
         definition.
 ```        
Closes # PB-1844

### CODE REVIEW

File: pkg/apis/v1/api.proto [hunk 1]
```diff
@@ -241,6 +241,16 @@ message BackupScheduleInfo {
     ObjectRef backup_location_ref = 19;
     ObjectRef pre_exec_rule_ref = 20;
     ObjectRef post_exec_rule_ref = 21;
+    // To get type of backup (generic or non-generic)
+    BackupType backup_type = 22;
+    message BackupType {
+        Type type = 1;
```
prashanthpx, line 247: Any specific reason to embed Type in a message? 
As BackupType is holding only one field, can we use the enum type directly? This will reduce one in-direction also.

`BackupType Type = 22` something like this

In most of our proto, we have followed the same.

siva-portworx: We use Invalid enum value as a common value in every enum definition. So we can not have the definition outside. We will get compilation error, if we have more than one enum value definition. That why we followed the extra level  of message definition in most of our previous enum definition.

prashanthpx (approved):

PR #112, opened 2021-09-17
pb-1888: added storage class map in restore object and create request.

siva-portworx:
pb-1888: added storage class map in restore object and create request.
Closes # pb-1888


ram-infrac (approved):
lgtm

PR #115, opened 2021-10-12
pb-1902 - swagger UI enabled with "authorize" tab

lalat-das:
In swagger-ui the user need to key-in bearer token which is enabled
via "authorize" button.

**What this PR does / why we need it**: This PR enables a "Authorize" button which enables user to provide TOKEN string so that further REST operations can be performed with proper permission.

Closes # pb-1902

### DIFF
--- a/pkg/apis/v1/api.proto
+++ b/pkg/apis/v1/api.proto
@@ -612,7 +612,7 @@ message EnumerateOptions {
 }
 
 service SchedulePolicy {
-    // Create creates new schedule policy.
+    // Creates new schedule policy.
     rpc Create(SchedulePolicyCreateRequest)
         returns (SchedulePolicyCreateResponse) {
         option (google.api.http) = {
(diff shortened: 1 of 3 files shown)


siva-portworx (approved):

pkg/apis/v1/api.proto
```diff
@@ -10,10 +10,28 @@ import "google/api/annotations.proto";
 import "google/protobuf/timestamp.proto";
 import "github.com/gogo/protobuf/gogoproto/gogo.proto";
 import "pkg/apis/v1/common.proto";
+import "protoc-gen-openapiv2/options/annotations.proto";
+//
+// import "protoc-gen-swagger/options/annotations.proto";
```
prashanthpx, line 15: Remove comented code

prashanthpx (approved):

pkg/apis/v1/api.proto
lalat-das: Done. Thanks

PR #120, opened 2021-11-19
pb-2038: Summary of the version API is corrected

lalat-das:
Altered the summary line to make it more sense in swagger json file.

Signed-off-by: Prashanth Kumar <prkumar@purestorage.com>

**What this PR does / why we need it**:Altered the summary line to make it more sense in swagger json file.

Closes #pb-2038


siva-portworx (approved):

pkg/apis/v1/api.swagger.json
```diff
@@ -2353,7 +2353,7 @@
     },
     "/v1/version": {
       "get": {
-        "summary": "Get gets the version of the server",
+        "summary": "gets the version of the server",
```
prashanthpx, line 2356: This might be applicable to other API's as well right? Might be later we need to look at holistically and fix all of them if applicable.

PR #124, opened 2022-02-24
pb-2178: proto changes for objectlock in backuplocation

lalat-das:
Added a bool variable in backuplocation to identify if the
object lock is enabled for this S3 bucket or not.

Signed-off-by: sivakumar subramani <ssubramani@purestorage.com>

**What this PR does / why we need it**: We need to use backuplocation object to be aware of object lock info of S3 bucket, hence created API to extract the same

Closes # pb-2178


pkg/apis/v1/api.proto
```diff
@@ -343,6 +343,7 @@ message BackupLocationInfo {
         Status status = 1;
         string reason = 2;
     }
+    bool objLockEnbled = 9;
```
siva-portworx, line 346: Here we use _ format instead of camel format for variable name. Please it.

siva-portworx (approved):

pkg/apis/v1/api.proto
lalat-das: ok

prashanthpx (commented):
One suggestion as there are a lot of vendor file changes, please have two separate commits.
1) One for proto changes 2) Second one for other vendor file updates.

PR #126, opened 2022-03-02
Changes to ManagedCluster API to accept region, return k8s version and cluster status

prashanthpx:
```
- Added placeholder constructs for GKE and Azure cloud providers
- Passing UID for cloud cred inspect apart from the name
- Additional field in ManagedClusterObject for k8s version and status of
  cluster is present in local datastore
```

baas-62


pkg/kubeauth/aws/aws.go
```diff
@@ -229,25 +228,29 @@ func GetRestConfigForAllClusters(awsConfig *api.AWSConfig) (map[string]*kubeauth
 		})
 		if err != nil {
 			logrus.Errorf("Failed to describe cluster %v: %v", *clusterName, err)
-			return nil, err
+			continue
```
siva-portworx, line 231: Is it not better to return error? If the he does not have describeCluster permission, we might be return empty list. That should not indicate to the user that he does not have cluster.

pkg/kubeauth/aws/aws.go
```diff
@@ -279,6 +282,7 @@ func getRestConfig(cluster *eks.Cluster, sess *session.Session) (*rest.Config, s
 	}
 	_, err = kubeauth.ValidateConfig(restConfig)
 	if err != nil {
+		logrus.Errorf("error validating kubeconfig: %v", err)
```
siva-portworx, line 285: Good to have the cluster name or some identity in the error message for debug.

siva-portworx (approved):

pkg/kubeauth/aws/aws.go
prashanthpx: Usually, if he is able to describe one cluster he should be able to others also from a permission point of view (as permission is at the IAM user level). Here the chances of error could be more on some issues with API for a particular cluster.  The thinking was if one Describe cluster fails, didn't want to throw an error or empty list as the user would be confused. 
We can get feedback from the test and Niranjan and see if this is ok or fail the operation itself.

PR #130, opened 2022-03-08
pb-2221: added auto-delete and ForObjectLock in schedule policy object

siva-portworx:
```
    pb-2221: added auto-delete and ForObjectLock in schedule policy object/

        - Removed the auto-delete from the backupschedule object
```
Closes # pb-2221


lalat-das (approved):
lgtm

PR #132, opened 2022-03-09
baas-63: Bulk addition of managed cluster and tracking cluster add st…

prashanthpx:
```
       - Adding new cluster status Success and Failure to track cluster
          status in case of cluster failure or success
        - Filtering cluster discovery to include only user-owned clusters
        - Addition of BulkAdd() API to add bulk managed clusters
```

baas-63:


pkg/apis/v1/api.proto
```diff
@@ -63,6 +63,12 @@ message ClusterInfo {
             Online = 1;
             Offline = 2;
             DeletePending = 3;
+            // Pending If a cluster addition is in pending state
+            Pending = 4;
+            // Failed If a cluster failed to be added
+            Failed = 5;
+            // Success If cluster was successfully added but not Online yet
+            Success = 6;
```
siva-portworx, line 71: If it is added successfully but online means, it should offline rt?

pkg/apis/v1/api.proto
```diff
@@ -1893,6 +1908,39 @@ message ManagedClusterInspectRequest {
...
+    Provider provider = 4;
+    oneof Config {
+        AWSConfig aws_config = 100;
+        GoogleConfig google_config = 101;
+        AzureConfig azure_config = 102;
+    }
+
+    message AWSConfig {
+        // Region for scanning clusters
+        string region = 1;
```
siva-portworx, line 1934: Are we planning to store the region in the cluster object. Why do we need region in bulk cluster add?

siva-portworx (approved):

pkg/apis/v1/api.proto
prashanthpx: Yes right, this sentence `added but not Online yet` conveys the same.

pkg/apis/v1/api.proto
prashanthpx: The region is not stored in cluster object, the above one is for `ManagedClusterBulkAddRequest`. As part of bulk add we need to fetch the kubeconfig internally (in px-backup) for which we need region.

pkg/apis/v1/api.proto
siva-portworx: If it is successfully added not online means, we may need to show that as Offline status instead of success status. So I think, success might not be a valid state itself.

PR #160, opened 2022-08-05
PB-2991: backup object to have cluster uuid along with name.

diptiranjanpx:
Signed-Off-By: Diptiranjan


pkg/apis/v1/api.proto
```diff
@@ -652,6 +653,9 @@ message EnumerateOptions {
     // set to nil and rest of the backupObject details will be returned in the
     // response.
     bool include_detailed_resources = 7;
+    // Filter to use with cluster uid on objects. Any object that contains the
+    // filter will be returned
```
siva-portworx, line 657: It would good, if you can add a comment to say that both name and uid needs to given together.

siva-portworx (approved):

prashanthpx (approved):

pkg/apis/v1/api.proto
diptiranjanpx: Backup enumerating can take both but lets say restore enumerating will still be requiring only cluster name as of now.

PR #163, opened 2022-10-14
pb-3159: Added nfs-share subpath and mount option

lalat-das:
Added nfs-share subpath. This directory will be created in nfs exported share directory. Mount option are passed during nfs file system.

Signed-off-by: Prashanth Kumar <prkumar@purestorage.com>

**What this PR does / why we need it**: We need to create  nfs fileshare sub path under which  backup related files will be created. Additionally mount option also scanned from user for passing during mounting the NFS FS.

Closes # pb-3159


pkg/apis/v1/api.proto
```diff
@@ -346,7 +346,11 @@ message BackupScheduleObject {
     BackupScheduleInfo backup_schedule_info = 2 [ (gogoproto.embed) = true ];
 }
 
-message NFSConfig { string nfs_server_addr = 1; }
+message NFSConfig {
```
siva-portworx, line 349: Nit: Since it is already NFSConfig struct, it is good to avoid nfs_ against each variable name.

lalat-das: Sure changed it

PR #173, opened 2023-01-17
PB-3406 : Add project mapping for rancher

px-kesavan:
Signed-off-by: Kesavan Thiruvenkadasamy <kthiruvenkadasamy@purestorage.com>

Closes #


pkg/apis/v1/api.proto
```diff
@@ -449,6 +449,8 @@ message BackupInfo {
     BackupShare backup_share = 28;
     BackupShare.AccessType user_backupshare_access = 29;
     ObjectRef cluster_ref = 30;
+    // Name of projects associated with resources
+    repeated string projects = 31;
```
siva-portworx, line 453: Does it make sense to have mention about rancher in the comment? So the use of the APIs will know it is specific to rancher project and not something generic.

pkg/apis/v1/api.proto
```diff
@@ -1565,6 +1568,8 @@ message RestoreCreateRequest {
     map<string, string> storage_class_mapping = 8;
     // backup object reference having name and id
     ObjectRef backup_ref = 9;
+    // mapping of the projects from the backup to which restore should be done
```
siva-portworx, line 1571: I feel it is better change the variable name also to include rancher in it.

siva-portworx: We might need to have a parameter in RestoreInfo def to store the project mapping details.

siva-portworx (approved):

pkg/apis/v1/api.proto
px-kesavan: Updated the comment and modified the variable to rancher_projects

pkg/apis/v1/api.proto
px-kesavan: fixed,
modified variable name to rancher_project_mapping
and added to RestoreInfo

PR #182, opened 2023-02-24
PB-3588

sgajawada1:
update namespace label selector field in backupschedule, backup, objects so that we can select namespace to be backup using labels provided by user.

Signed-off-by: Santosh Kumar Gajawada [sgajawada@purestorage.com](mailto:sgajawada@purestorage.com)

Change the NS Label Selector type to string instead of a map
Fix static check issue 

Closes #PB-3588


siva-portworx (approved):

lalat-das:
Shouldn't this patch should go to master branch also ..

siva-portworx:
yes, it has to

PR #202, opened 2023-06-30
PB-3980: API Changes to add filter as owner UID

pallav-px:

Closes #


pkg/apis/v1/api.proto
```diff
@@ -735,6 +735,9 @@ message EnumerateOptions {
     // Currently backup enumerate request should pass this along with cluster
     // name.
     string cluster_uid_filter = 8;
+    // Filter to use owner uids on objects. Any object that contains the
+    // filter will be returned
+    repeated string owner = 9;
```
lalat-das, line 740: make the varibale name a plural name.. as it is going to hold more number of owners..

lalat-das (approved):
just make the variable name plural.. rest LGTM

pkg/apis/v1/api.proto
pallav-px: done

PR #209, opened 2023-07-26
PB-4101: Non RBAC Objects to have object UID in Request APIs

pallav-px:

This PR is need to add object UID field in all non RBAC object such as backup, backupSchedule, restore and cluster to have a object UID field in their request field to get the unique object while inspecting and deleting the particular object. Since after px-admin-group changes. admin can see all the objects of all users. So its difficult to fetch objects uniquely with only name.

**Which issue(s) this PR fixes** 
Its fixes the issue when admin wasn't able to get the exact object of particular user with its only name in non rbac object cases


prashanthpx:
@pallav-px You need to add {uid} in the http path for inspect for REST API's

lalat-das (approved):
lgtm

PR #210, opened 2023-07-27
PB-4114: Proto Changes to Add Cluster Ref on non rbac objects

pallav-px:

This PR need so that we can pick the exact object it belongs to which cluster in case of multiple cluster with same name

This PR will pix picking the object with exact cluster object in case of inspect and delete of non rbac objects.

Closes #


siva-portworx (approved):

prashanthpx (commented):
LGTM

PR #223, opened 2024-01-17
PB-5291: Kubevirt - Phase 2 Implement gRPC changes for VM Backup

vsundarraj-px:
This PR will implement additional field to support Kubevirt phase 2 

Closes # 
PB-5291

Unit tested by building stork/ px-backup to see there is no build failure and ensure the existing functionality is broken by this PR.


pkg/apis/v1/api.proto
```diff
@@ -344,6 +348,10 @@ message BackupScheduleInfo {
     // Target namespace of CR
     string target_namespace = 24;
     ObjectRef cluster_ref = 25;
+    // Is this a VM Backup object
+    BackupObjectType backup_object_type = 26;
+    // Skip auto rules for VM Backup Object
+    bool skipVMAutoExecRules = 27;
```
dbinnal-px, line 354: Just thinking about the naming convention that is being followed over.
```suggestion
    bool skip_vm_auto_exec_rules = 27;
```

dbinnal-px (approved):

px-kesavan (approved):

pkg/apis/v1/api.proto
```diff
@@ -988,6 +1007,10 @@ message BackupScheduleUpdateRequest {
     ObjectRef backup_location_ref = 14;
     ObjectRef pre_exec_rule_ref = 15;
     ObjectRef post_exec_rule_ref = 16;
+    // Is this a VM Backup object
+    BackupObjectType backup_object_type = 17;
```
siva-portworx, line 1011: Are we planning to change the VM backup object of a schedule? I means initially default and then updating it to be VMType?
Initially, when it was default, he might have selected the includeObject ( custom selection). So I would suggest to check with Niranjan once.

pkg/apis/v1/api.proto
siva-portworx: Yes, let us please change it in all other places as well.

pkg/apis/v1/api.proto
```diff
@@ -230,6 +230,10 @@ message CloudCredentialObject {
     CloudCredentialInfo cloud_credential_info = 2;
 }
 
+enum BackupObjectType {
+        default = 0; //default
+        virtualMachine = 1;
```
siva-portworx, line 235: Can we add some comments for the virtualMachine. I would suggest to have some comment for the enum ( why the type really means)as these comments will come in the Swagger as well.

pkg/apis/v1/api.proto
vsundarraj-px: You are right we do not want to allow changing this as the support is largely from UI not from cli.

pkg/apis/v1/api.proto
vsundarraj-px: Added

pkg/apis/v1/api.proto
vsundarraj-px: updated

pkg/apis/v1/api.proto
```diff
@@ -230,6 +230,11 @@ message CloudCredentialObject {
     CloudCredentialInfo cloud_credential_info = 2;
 }
 
+// defines if backupObjectType is default for all namespace  or  virtualMachine for kubevirt VirtualMachine specific backup
+enum BackupObjectType {
+        default = 0; //default for all namespace backup
```
prashanthpx, line 235: We should avoid using 0 for our enum values as default value of enum from proto lanaguge is zero. In all our PX repos we follow this practice.
Ref link https://protobuf.dev/programming-guides/dos-donts/

prashanthpx: Also enum name should start with a caps

siva-portworx (approved):
LGTM with comment on Enum is fixed.

pkg/apis/v1/api.proto
vsundarraj-px: Could not use Invalid as BackupObjectType is declared global, and Invalid is already used in common.proto.
Hence using Unused.

PR #225, opened 2024-01-24
PB-5431 : Refining Backup : Changes to add CSISnapshotClass Mapping

pallav-px:
**What this PR does / why we need it**: This PR add schema for CSI Snapshot Class Mapping for Provisioner to their respective volumeSnapshotClass. This will increase flexibility to for user to select provisioner based Volume Snapshot Class so that there will no confusion which PVC will be treated as CSI, KDMP or PXD

Closes #


dbinnal-px (approved):

vikasit12 (approved):
LGTM

pkg/apis/v1/api.proto
```diff
@@ -1624,6 +1625,8 @@ message BackupCreateRequest {
     BackupObjectType backup_object_type = 17;
     // Skip auto rules for VirtualMachine Backup Object type
     bool skip_vm_auto_exec_rules = 18;
+    // voume snapshot class mapping for csi based backup <provisioner, volumesnapshotclass>
```
siva-portworx, line 1628: typo voume

siva-portworx: Also can we add (optional) in the comment

siva-portworx: Also, it is good to document the string that we are planning to use for picking up the default VSC here. So that if some try with API way, they are aware of the string value for default.

pallav-px: Sure

PR #231, opened 2024-02-01
pb-4263 : Email Alert 

t-nidhin:
This PR is the merge with the new feature Email Alert.
Following are some of the explanations for stats added in messages
**Reciever Object:**
1. Metadata
2. ReceiverInfo:
- Type ( Currently receiver is of type email, in future this can have slack or pager, which are supported in alertmanager)
- config: (Emailconfig)
1. from: from address used in SMTP server
2. host: host address of the SMTP server
3. port: SMTP port, which is default 25
4. encryption SSL: This is additional param added since keycloak requires this param
5. encryption start TLS: param used in keycloak 
6. authentication: this flag is set, if the smtp server has authentication
7. auth_username: username for the server
8. auth_password: password for the server

**Recipient Object**
1. Metadata
2. RecipientInfo:
- Again type is referred here, this should be same as parent type receiver
- recipient_id: This is the list of id's, in this release this would be email
- active: param to check if the current user has set the email notification to enabled/disabled.
- objectRef: reference to parent object Receiver
- Severity: Type of email verity for the recipients to be sent.

Closes #
[PB-5305](https://portworx.atlassian.net/browse/PB-5305)
[pb-4945](https://portworx.atlassian.net/browse/PB-4945)
[pb-4403](https://portworx.atlassian.net/browse/PB-4403)

[PB-5305]: https://portworx.atlassian.net/browse/PB-5305?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/apis/v1/api.proto
```diff
@@ -1367,6 +1367,276 @@ message CloudCredentialOwnershipUpdateRequest {
...
+    // Sender email address
+    string from = 1;
+    // receiver email, User level set of receiver emails list
+    // repeated ObjectRef recipient_ref = 2;
+    // Host IP Address
+    string host = 3;
+    // Host port default value is 25
+    string port = 4;
+    // Encryption if any (optional) default value is false
+    bool encryption_ssl = 6;
```
dbinnal-px, line 1433: Just a small observation on the field number ordering. Same with the above field string host = 3;
```suggestion
    bool encryption_ssl = 5;
```

pkg/apis/v1/api.proto
```diff
@@ -1367,6 +1367,276 @@ message CloudCredentialOwnershipUpdateRequest {
...
+            body: "*"
+        };
+    }
+}
+
+message EmailConfig {
+    // Sender email address
+    string from = 1;
+    // receiver email, User level set of receiver emails list
+    // repeated ObjectRef recipient_ref = 2;
```
dbinnal-px, line 1427: Just wondering why this is being commented out?

pkg/apis/v1/api.proto
t-nidhin: Good eye, Thank you.

dbinnal-px (approved):
LGTM

pkg/apis/v1/api.proto
```diff
@@ -1367,6 +1367,274 @@ message CloudCredentialOwnershipUpdateRequest {
...
+            option (google.api.http) = {
+                get : "/v1/receiver/{org_id}"
+            };
+        }
+
+    // Returns the details of a specific alert manager receiver
+    rpc Inspect(ReceiverInspectRequest)
+        returns (ReceiverInspectResponse) {
+            option (google.api.http) = {
+            get : "/v1/receiver/{name}/{uid}"
```
prashanthpx, line 1392: We aslo need to add `include_secrets` parameter to the REST API. I know for existing API's it's not added but lets try to add for this.

prashanthpx: Also add `type`

t-nidhin: done, addressed.

pkg/apis/v1/api.proto
```diff
@@ -1367,6 +1367,274 @@ message CloudCredentialOwnershipUpdateRequest {
 
 message CloudCredentialOwnershipUpdateResponse {}
 
+service Receiver {
```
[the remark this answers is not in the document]
t-nidhin: Valid concern, but I named it in conjunction with the Prometheus v1 api's. The change now will take some time.

PR #235, opened 2024-02-25
PB-5834: API Changes ->  DIRECT_KDMP for Backup Object Level

pallav-px:

Closes #


dbinnal-px (approved):

pkg/apis/v1/api.proto
```diff
@@ -351,6 +351,8 @@ message BackupScheduleInfo {
     bool skip_vm_auto_exec_rules = 27;
     // volume snapshot class mapping for csi based backup <provisioner(string), volumesnapshotclass(string)> (optional)
     map<string, string> volume_snapshot_class_mapping = 28;
+    // option to take backup as direct kdmp 
+    bool direct_kdmp = 29;
```
prashanthpx, line 355: Any reason we need this to be part of backupschedule object as backup driver info already has driver type.

pkg/apis/v1/api.proto
```diff
@@ -534,6 +536,8 @@ message BackupInfo {
     bool skip_vm_auto_exec_rules = 36;
     // volume snapshot class mapping for csi based backup <provisioner(string), volumesnapshotclass(string)> (optional)
     map<string, string> volume_snapshot_class_mapping = 37;
+    // option to take backup as direct kdmp 
+    bool direct_kdmp = 38;
```
prashanthpx, line 540: Same as above

pkg/apis/v1/api.proto
```diff
@@ -1052,6 +1059,8 @@ message BackupScheduleUpdateRequest {
     ObjectRef post_exec_rule_ref = 16;
     // volume snapshot class mapping for csi based backup <provisioner(string), volumesnapshotclass(string)> (optional)
     map<string, string> volume_snapshot_class_mapping = 17;
+    // option to take backup as direct kdmp 
+    bool direct_kdmp = 18;
```
prashanthpx, line 1063: One confusion which might cause here is if a user was on PX vol and had incremental backups and in between he switches to direct KDMP then it might lead to confusing as what incremental mean now where already we have confusion with generic backup. I am thinking if a user wants to switch, we should ask them to create new schedule.

pkg/apis/v1/api.proto
pallav-px: BackupSchedule gives values to Backup. So for per backup level direct kdmp we need to add in BackupSchdeuleInfo too

pkg/apis/v1/api.proto
pallav-px: Now its per backup level direct kdmp we should store the value to understand driver easily

pkg/apis/v1/api.proto
pallav-px: If we change the driver for backup schedule now. Each new backup after change will be full backup and those previous incremental backup will be irrelevant . This should be documented

siva-portworx (approved):

PR #243, opened 2024-05-08
pb-6910: preserve uid & gid with which volume is backed up

lalat-das:
During backup due to restricted PSA the app will use certain uid and gid. Same uid Gid need to be used to restore while using a JOB-POD.

**What this PR does / why we need it**: During backup due to restricted PSA the app will use certain uid and gid. Same uid Gid need to be used to restore while using a JOB-POD.

Closes # pb-6910


pkg/apis/v1/api.proto
```diff
@@ -585,6 +585,10 @@ message BackupInfo {
         string pvc_id = 12;
         string provisioner = 13;
         string volumesnapshot = 14;
+        // if as part of restrictive PSA setting, an UID or GID is used
+        // during backup then same value should be used during restore operation
```
vsundarraj-px, line 589: Isnt this mostly for inplace restore to same namespace on same cluster? How does this apply for new cluster ? UID and GID would not be same.

pkg/apis/v1/api.proto
```diff
@@ -568,6 +568,12 @@ message BackupInfo {
         string name = 2;
     }
 
+    message VolumeSecurityInfo {
```
vsundarraj-px, line 571: This is miss leading name as this is not the security IDs for a volume but for Job Pods that we spin up to access these volumes. Volumes as such do not have any UID/GID specification. So lets use the name that does not create this confusion.

Also please comment where this will be used and for reason. A story is merrier to me as this is very unique field used for specific processing for a specific security configuration.

vsundarraj-px: I would suggest to keep SecurityContext and can use Suffix or Prefix as necessary. One name I can think of is
KDMPJobSecurityContext

pkg/apis/v1/api.proto
```diff
@@ -568,6 +568,12 @@ message BackupInfo {
...
+        // if as part of restrictive PSA setting, an UID or GID is used during backup then same value should be used during restore operation (optional)
+        int64 uid = 1;
```
vsundarraj-px, line 573: Please use runAsUser itself so that we know what is the correlation within in k8s

pkg/apis/v1/api.proto
```diff
@@ -568,6 +568,12 @@ message BackupInfo {
...
+        int64 gid = 2;
```
vsundarraj-px, line 574: Similarly runAsGroup?

pkg/apis/v1/api.swagger.json
```diff
@@ -3435,6 +3435,23 @@
         },
         "volumesnapshot": {
           "type": "string"
+        },
+        "volume_security_info": {
+          "$ref": "#/definitions/BackupInfoVolumeSecurityInfo"
+        }
+      }
+    },
+    "BackupInfoVolumeSecurityInfo": {
```
vsundarraj-px, line 3444: You need to move the definition of the message with in the Volume to get Specific nestedName from proto parsing.

pkg/apis/v1/api.proto
lalat-das: Sure make sense...

pkg/apis/v1/api.proto
lalat-das: Sure I will rename it as suggested.

pkg/apis/v1/api.proto
```diff
@@ -587,6 +587,11 @@ message BackupInfo {
         string pvc_id = 12;
         string provisioner = 13;
         string volumesnapshot = 14;
+        // As part of restrictive PSA setting, if an UID or GID is used during backup then same value should be used during restore (optional)
+        message JobSecurityContext {
```
vsundarraj-px, line 591: Is the decleration of this definition missed?

lalat-das: Added the member defining the structure.

vsundarraj-px (approved):
LGTM

PR #244, opened 2024-05-10
pb-6810: Added resource group name to S3Config of the backuplocation object

siva-portworx:
```
pb-6810: Added resource group name to S3Config of the backuplocation object
```
Closes # pb-6810

N.A
To azure SDK call need reource group name to get the retention period for the locked bucket. 
We will be getting this from user during the backuplocation creation.


pkg/apis/v1/api.proto
```diff
@@ -165,6 +165,9 @@ message S3Config {
     // Azure environment type
     // Currently supporting only AZURE_GLOBAL and AZURE_CHINA
     AzureEnvironmentType azure_environment = 7;
+    // Azure resource group name.
+    // This is needed in the case of azure immutable bucket to get the retention period
```
lalat-das, line 169: Please make sure to add (optional) in the comment if it is an optional argument in BL related ops

lalat-das (approved):
lgtm, other than that optional related comment.

PR #251, opened 2024-08-04
PB-7669: adding retention time stamp

dbinnal-px:
Adding new field for retention timestamp

Closes #
https://purestorage.atlassian.net/browse/PB-7669


pkg/apis/v1/api.proto
```diff
@@ -556,6 +556,7 @@ message BackupInfo {
     map<string, string> volume_snapshot_class_mapping = 37;
     // option to take backup as direct kdmp 
     bool direct_kdmp = 38;
+    google.protobuf.Timestamp retention_time = 39;
```
siva-portworx, line 559: Add a comment that this will store the timestamp, when the retention period of locked backup will be expired.

siva-portworx (approved):

aks-px (approved):

vsundarraj-px (approved):
Agree with @siva-portworx please add comment.

PR #252, opened 2024-08-09
feat: PB-7779 implement cluster delete to support cluster share feature

vsundarraj-px:
This PR add two new flags to ClusterInfo deleted_by and delete_all_cluster_backup.
WIth cluster share we allow cluster deletion by cluster owner and also SuperAdmin. This deleted_by will identify whose backups should we delete while deleting the cluster with option deleteBackups. 

Similarly delete_all_cluster_backups field is introduced to cleanup backups created by all shared users including the other super admin user. this field is allowed to be executed only by super admin. 
 
Closes #
PB-7779

Unit testing results will be shared with px-backup PR


t-nidhin (approved):

pkg/apis/v1/api.proto
```diff
@@ -1323,6 +1329,9 @@ message ClusterDeleteRequest {
     bool delete_restores = 4;
     // indicate the uid of the particular object
     string uid = 5;
+    // delete_all_cluster_backup if set will delete all backups on the cluster
+    // allowed only to super admin Role
```
siva-portworx, line 1333: I feel, we need to change the comment of the delete_backups as well. Also comment looks more generic for delete_all_cluster_backup. I pefer,

```
// delete_backups indicates whether all the backup created by cluster owner and belong to the cluster need to 
  // be deleted or retained. This can be set, when cluster delete was trigged by user with 
  // any role
  bool delete_backups = 3;
  
   // delete_all_cluster_backups  indicates whether all the backup owned by cluster owner and belong to the 
   // cluster need to be deleted or retained. This will be set only, when cluster delete was trigged by user with 
   // SuperAdmin role
  
    bool delete_all_cluster_backups = 6;
```

siva-portworx (approved):

pkg/apis/v1/api.proto
vsundarraj-px: fixed

prashanthpx: Some thougths 
1. Are we going to handle the case when both are set? What would be the behaviour?
2. If a superadim creates the cluster, can he use any of the flags?
3.  When a super_admin is using the `delete_all_cluster_backups ` flag, should we show warning msg that other users also have created the backup and that would be deleted? Just a warning to the super_admin but not stop him from deleting.

siva-portworx: **delete_backups** - this will be used by any user ( superAdmin or non superAdmin user). With this flag, we will delete only the backup created by the user and that belongs to the cluster.

**delete_all_cluster_backups** - this will be used by only SuperAdmin user. With this flag we will delete all the backup that belongs to the cluster, even if it is not created by the user who initiates the delete. We will return error, if non-admin user set the delete_all_cluster_backups. If Super Admin set both the flags, delete_all_cluster_backups will take priority. 

Also warning can be given only in the UI and in the UI, currently these flags are not used. 
Warning in the API is not possible. So for API, we should document the purpose of the flags clearly.

vsundarraj-px: Agree with @siva-portworx. This is not supported from UI as of now. Only from API and more over this is not default. It is something that super admin uses it cauteously. Because for default its false and for deletion of his own backups is delete_backups which has clear swagger message and further on the delete_all_cluster_backups is explicit flag allowed only to super admin. With all these I dont think we should have warnings.

prashanthpx: As we know from UI it will never be used, taking a step back I m thinking do we need to really support this (even the older one). It might be good time now to clean it up. As we don't know who is using it(I doubt anyone is using it) we can always call out this is deprecated. Lot of engineering to support which is hardly used

PR #256, opened 2024-08-22
PB-7774: Adding proto changes for addition of timestamps

kgarg-px:
Introducing timestamps for backup objects and their resources 
Closes #


pkg/apis/v1/api.proto
```diff
@@ -558,6 +558,12 @@ message BackupInfo {
     bool direct_kdmp = 38;
     // this will store the timestamp, when the retention period of locked backup will be expired
     google.protobuf.Timestamp retention_time = 39;
+    // This will store timestamp for the completion of volumes backup
+    google.protobuf.Timestamp volumes_completion_time = 40;
+    // This will store timestamp for the completion of resources backup
+    google.protobuf.Timestamp resources_completion_time = 41;
+    // This will store the total completion time of the entire backup
+    google.protobuf.Timestamp total_completion_time = 42;
```
dbinnal-px, line 566: just a thought, can we name it as backup_completion_time or just completion_time which would more convinient

dbinnal-px (approved):

aks-px (approved):

pkg/apis/v1/api.proto
```diff
@@ -571,6 +571,12 @@ message BackupInfo {
     bool direct_kdmp = 38;
     // this will store the timestamp, when the retention period of locked backup will be expired
     google.protobuf.Timestamp retention_time = 39;
+        // This will store timestamp for the completion of volumes backup
+    google.protobuf.Timestamp volumes_completion_time = 40;
+    // This will store timestamp for the completion of resources backup
+    google.protobuf.Timestamp resources_completion_time = 41;
+    // This will store the total completion time of the entire backup
+    google.protobuf.Timestamp total_completion_time = 42;
```
pallav-px, line 579: can we make these field generic for all object. In case if we need restore also in future we needed that struct to add and not all fields

vikasit12: I second this thought the 3 variables can be made as different struct and can be used in restore too in future

siva-portworx (approved):

PR #258, opened 2024-08-30
PB-7615:  Adding isSyncedBackup flag to identify the synced backups

shkumari-px:
**What this PR does / why we need it**: it adds a flag isSyncedBackup to check if the backup is a synced backup

Closes # https://purestorage.atlassian.net/browse/PB-7615

**Special notes for your reviewer**: na


Makefile
```diff
@@ -64,7 +64,7 @@ vet:
 	go vet $(PKGS)
 
 staticcheck:
-	GOFLAGS="" go install honnef.co/go/tools/cmd/staticcheck@v0.4.7
+	GOFLAGS="" go install honnef.co/go/tools/cmd/staticcheck@latest
```
dbinnal-px, line 67: any reason why we are changing this? Because as per I remember when it was 'latest' it was giving error while building docker image due to which we had given a specific version.

pkg/apis/v1/api.proto
```diff
@@ -2062,6 +2064,8 @@ message BackupCreateRequest {
     map<string, string> volume_snapshot_class_mapping = 19;
     // option to take backup as direct kdmp 
     bool direct_kdmp = 20;
+    // this flag will check if the backup is a synced backup
+    bool isSyncedBackup = 21;
```
dbinnal-px, line 2068: I guess we don't need this in create request. As a synced backup is created through the backupsync

Makefile
kgarg-px: +1 please confirm if we have to use latest or 0.4.7

pkg/apis/v1/api.proto
kgarg-px: +1 having the bool in `BackupInfo` is enough, Since backup sync objects are created through BackupSync function and not backupCreate function in px-backup. Having the variable in BackupInfo as a property of backup (BackupInfo message) should suffice here.

siva-portworx:
@shkumari-px Let us add a flag called backup_variant ( generic backup type) and add enums to classify them as manual /schedule /synced backup.
cc @prashanthpx

Makefile
shkumari-px: Screenshot 2024-09-02 at 12 07 44 PM
Go version 1.23 with staticcheck version 0.4.7 still gives error

pkg/apis/v1/api.proto
shkumari-px: @dbinnal-px @kgarg-px 
When we sync the backup, we create a new backup object in the mongodb
so you need the flag in the createBackupRequest as well

shkumari-px:
Closing this pr
If we have a synced backup then the clusterRef will be nil for it
This check can be used to verify if it’s a synced backup

PR #273, opened 2025-01-06
PB-9235: Added proto changes for supporting smtp tls

px-kesavan:
**What this PR does / why we need it**: Added proto changes for supporting input of certs required for smtp

Closes # https://purestorage.atlassian.net/browse/PB-9235


pkg/apis/v1/api.proto
```diff
@@ -1578,6 +1578,23 @@ message EmailConfig {
         (gogoproto.jsontag) = "encryptionkey",
         (gogoproto.moretags) = "secure:\"true\""
     ];
+    // Use custom certificate for SMTP server
+    bool use_custom_certificate = 9;
+    // Certificate file content
+    string cert_file = 10 [
```
siva-portworx, line 1584: It would be better to create a message def for ca cert in common.go and use it where ever required.

px-kesavan: I have moved to cert content under tlsConfig , placed in common.proto. PTAL

t-nidhin (approved):

siva-portworx (approved):

PR #284, opened 2025-03-20
PB-10126: added advance_resource_label_selector to make the resourcelabel to be string type instead of map

siva-portworx:
```
PB-10126: added advance_resource_label_selector to make the resourcelabel to be string type instead of map.
With map, it is difficult to pass the operator values "=" in "a=b" or "!=" in "c!=d". So now make label as string type, which will be similar to namespace label.
```
Closes # PB-10126


pkg/apis/v1/api.proto
```diff
@@ -1084,6 +1088,8 @@ message BackupScheduleCreateRequest {
     bool parallel_backup = 25;
     // option to enable to keep the CR status of the resources in the backupschedule.
     bool keep_cr_status = 26;
+    // older label selector does not support string type. Map type has contraint of supporting operator, So introducing this new parameter
+    string advance_resource_label_selector = 45;
```
kravichandran-px, line 1092: minor: 45 to `27`?

shkumari-px (approved):

kravichandran-px (approved):

PR #289, opened 2025-04-16
PB-10326 : Add identifiers field to CreateMetadata and BackupScheduleDeleteRequest for bulk update and delete of backupSchedule

vikasit12:
- Modified CreateMetadata to support repeated ObjectRef identifier
- Modified BackupScheduleUpdateRequest to add BackupObjectType backup_object_type
- Modified BackupScheduleDeleteRequest to add repeated ObjectRef identifier and BackupObjectType backup_object_type
- Delete RPC: additional POST binding /v1/backupschedule/{org_id}/delete
- Modified the Delete RPC in BackupSchedule service to require 'name' as a path parameter alongside 'org_id' and 'uid'.
- Updated the delete/update message to include repeated fields for 'policy_ref' and 'exclude_object' for enhanced filtering capabilities.

Enables PX-Backup to support bulk delete and update in backup schedule

Closes #


pkg/apis/v1/api.proto
```diff
@@ -1147,6 +1151,17 @@ message BackupScheduleUpdateRequest {
     bool parallel_backup = 19;
     // older label selector does not support string type. Map type has contraint of supporting operator, So introducing this new parameter
     string advanced_resource_label_selector = 20;
+    // backup object type is all namespace or virtualMachine specific backup
+    BackupObjectType backup_object_type = 21;
```
vsundarraj-px, line 1155: why do we need backup_object_type in update request?

vikasit12: > why do we need backup_object_type in update request?

@vsundarraj-px 
As mentioned in the scope in our design document

Supporting selection of all backupSchedule as per view basis(NS/VM)
Supporting selection of substring based backupSchedule as per view basis(NS/VM)

I need to find the selection (All/Substring) was passed on by user from which view and i need to work on that set of view only.

kravichandran-px (approved):

pkg/apis/v1/api.proto
```diff
@@ -1194,6 +1209,21 @@ message BackupScheduleDeleteRequest {
     // delete_backups is deprecated, not to be used.
     bool delete_backups = 3;
     string uid = 4;
+    //  ObjectRef can contain:
+    // - specific list of name+uid (e.g., 10 items)
+    // - wildcard: name = "*" or name = "pxb-*"
+    repeated ObjectRef identifiers = 5;
```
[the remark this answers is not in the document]
vikasit12: > Are we planing to support all type of search operators? a*, a?, a+....

Name of backupSchedule doesn't support these identifiers.

As per design document,
As a user, I can select specific backup schedules by name and UID, or select all ("") or filter-based selections (e.g., "pxb-").
UI will append in * and send to backend so that we know that its a search/enumerate based operation.

pkg/apis/v1/api.proto
```diff
@@ -1194,6 +1209,21 @@ message BackupScheduleDeleteRequest {
...
+    // backup object type is all namespace or virtualMachine specific backup
+    BackupObjectType backup_object_type = 6;
+
+        message BackupObjectType {
+        Type type = 1;
+        enum Type {
+            Invalid = 0;
```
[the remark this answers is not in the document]
vikasit12: > Should have VM, NS, All?

Actually it should have but as historically we have been using this and to maintain uniformity for someone trying to automate I have kept it as it is used everywhere else in proto
https://github.com/portworx/px-backup-api/blob/master/pkg/apis/v1/api.proto#L440-L447
https://github.com/portworx/px-backup-api/blob/master/pkg/apis/v1/api.proto#L673-L680
https://github.com/portworx/px-backup-api/blob/master/pkg/apis/v1/api.proto#L817-L824
https://github.com/portworx/px-backup-api/blob/master/pkg/apis/v1/api.proto#L1100-L1107
https://github.com/portworx/px-backup-api/blob/master/pkg/apis/v1/api.proto#L2117-L2124
https://github.com/portworx/px-backup-api/blob/master/pkg/apis/v1/api.proto#L2262-L2269

We may change this but this may cause confusion for user

pkg/apis/v1/api.proto
```diff
@@ -1039,7 +1039,11 @@ service BackupSchedule {
     rpc Delete(BackupScheduleDeleteRequest)
         returns (BackupScheduleDeleteResponse) {
         option (google.api.http) = {
-            delete : "/v1/backupschedule/{org_id}/{name}"
+            delete: "/v1/backupschedule/{org_id}/{uid}"
```
pallav-px, line 1042: I hope changing required fields won't be an issue for customers who are already using in their automation

vikasit12: @pallav-px it will be but this is required as after superadmin this might become an issue going forward.

prashanthpx (approved):

pkg/apis/v1/api.proto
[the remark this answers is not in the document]
siva-portworx: Here we are giving option to the customer, such that in which group of backupschedule, we want to search.
only VM type, only NS type or Both. With the current option, I am not sure how will you achieve it?

pkg/apis/v1/api.proto
vikasit12: sure @prashanthpx

pkg/apis/v1/api.proto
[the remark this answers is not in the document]
vikasit12:
```
	// Construct matchField for filtering by backup_object_type if provided
	matchField := make(map[string]interface{})
	if objType := req.GetBackupObjectType(); objType != nil &&
		objType.GetType() != api.BackupScheduleDeleteRequest_BackupObjectType_Type(api.BackupScheduleInfo_BackupObjectType_Invalid) {
		matchField[datastore.BackupScheduleInfoBackupObjectTypeKey] = objType.GetType().String()
	}
	
	// Create FilterKeyValue to be used by internal getFilter()
	filterKV := &datastore.FilterKeyValue{
		Value: &datastore.FilterValue{
			MatchField: matchField,
		},
	}
```
	
	Using the matchFields of our already defined fitlerKV, we will be achieving it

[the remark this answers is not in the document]
siva-portworx: As discuss, let us not us Invalid value as default both option. For any valid option, we need to have a valid enum definition.

pkg/apis/v1/api.proto
```diff
@@ -1153,6 +1153,10 @@ message BackupScheduleUpdateRequest {
     string advanced_resource_label_selector = 20;
     // backup object type is all namespace or virtualMachine specific backup
     BackupObjectType backup_object_type = 21;
+    // multiple policy reference for filtering
+    repeated ObjectRef policy_ref = 22;
+    // multiple name + uid for excluding
+    repeated ObjectRef exclude_object = 23;
```
siva-portworx, line 1159: This exclude_object is the list of backupschedule object that we want to avoid rt?
This will be applicable when policy_ref is present or not rt? 
In that case, it is better, we mention that details in the comment.

siva-portworx: Nit: It better to name it as exclude_object --> exclude_objects as it is an array.

siva-portworx (approved):

pkg/apis/v1/api.proto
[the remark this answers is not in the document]
vikasit12: @siva-portworx as per discussion with Anjan he wants UI and proto to be in sync so not putting both option for now as UI doesn't display/provide this option.

pkg/apis/v1/api.proto
vikasit12: @siva-portworx This exclude_object doesn't depend directly on policy_ref, it depends on identifiers. If and only if wild card (*) is present with sub-string or select All then this filter has value. I am updating the name and comment

Scenario                                                                           Result
Only wildcard identifiers, no exclude                            OK
Wildcard identifiers + exclude object                           OK
Exact match identifiers + exclude object                     Error
Missing name/uid in policy_ref                                      Error
Missing name/uid in exclude_object                             Error
Mixing wildcard and exact matches in identifiers        Error
No identifiers, fallback to single name+uid fields        OK

PR #290, opened 2025-04-18
PB-10285 implement retry failed VM for failed and partial success VM …

vsundarraj-px:
…backups

This PR exposes new gRPC call to retry failed VM backup
Closes #


pkg/apis/v1/api.proto
```diff
@@ -2188,6 +2196,19 @@ message BackupShareUpdateRequest {
     string uid = 4;
 }
 
+// Request message for retrying failed backup
+message BackupRetryRequest {
+    string org_id = 1;
+    string name = 2;
+    string uid = 3;
+    repeated ResourceInfo include_resources = 4;
+    google.protobuf.BoolValue skip_vm_auto_exec_rules = 5;
```
kravichandran-px, line 2205: @vsundarraj-px You had mentioned about having a `namespaces` field that compliments `include_resources`. Just double checking if we are not going ahead with that. Im all in not have it, to avoid providing too many fields/options :)

vikasit12, line 2205: Is skip_vm_auto_exec_rules an optional field?
Why are we specifically using `google.protobuf.BoolValue` rather than `bool`
Are we suggesting skip_vm_auto_exec_rules can be null too?

kravichandran-px: In retry case, hope this field skips not just auto rules but the custom rules as well if set?

vsundarraj-px: I have updated the design document. Namespace is overhead specially when VMs are already provided with namespace. Hence removed.

vsundarraj-px: @vikasit12 the reason boolValue is used instead of bool with bool the default value is false. if the original backup was set to true.. how do we know if the value in the retryRequest was default value or the user specifically set it to false.. Hence I had to use boolValue. If no value is sent then this field will be nil and we will use the value that is originally set on the backup. 

Custom rule will be copied in case retryRequest does not provide any value. However, if customer changes to false to true, retry does not support custom rule.

kravichandran-px (approved):

pkg/apis/v1/api.proto
```diff
@@ -2188,6 +2197,27 @@ message BackupShareUpdateRequest {
...
+    repeated ResourceInfo include_resources = 4;
+    // Modify Skip auto rules for VirtualMachine Backup Object type
+    // This is only applicable for VirtualMachine backup object type
+    // Will update only if the orignal backup value differ. If original backup is set to true
+    // and if this field is set to true then no change is applied while retry. 
+    google.protobuf.BoolValue skip_vm_auto_exec_rules = 5;
+}
+
+// Response message for BackupRetryRequest
+message BackupRetryResponse {}
```
siva-portworx, line 2218: Can we return the newly created backup object, so that call will know the new name and the new UID of the backupobject?

PR #293, opened 2025-04-24
PB-10525: API Changes to support Sort Options for Enumerate Objects

pallav-px:

This PR adds support for sorting objects in the EnumerateOptions request.
It introduces a new repeated field sort_fields which allows users to specify one or more fields to sort on, with an option to define the order (ascending or descending).

Closes #


pkg/apis/v1/api.proto
```diff
@@ -846,6 +846,20 @@ message TimeRange {
...
 
+// Sorting structure
+message SortOption {
+    // Name of the field to sort by
+    string field = 1;
+    // Sort order: ASCENDING or DESCENDING
+    SortOrder order = 2;
+    // Enum for sort order
+    enum SortOrder {
+        ASCENDING = 0;
```
lalat-das, line 857: Shouldn't the value be like directly mapped to mongo's ascending descending values ? That way you won't have to map these numbers to corresponding mongo numbers.
1 for ascending order.
-1 for descending order.

0 should represent someone didn't fill anything to make it backward compatible.

pkg/apis/v1/api.swagger.json
```diff
@@ -198,6 +198,25 @@
...
+            "name": "enumerate_options.sort_by.order",
+            "description": "Sort order: ASCENDING or DESCENDING.",
+            "in": "query",
+            "required": false,
+            "type": "string",
+            "enum": [
+              "ASCENDING",
+              "DESCENDING"
+            ],
+            "default": "ASCENDING"
```
lalat-das, line 219: default should be descending latest timestamp should be the first followed by lesser numbers ...  this will keep the older behavior intact. check mongodb.go file

if filterKV.Value.SortOption == 0 {
       // Default latest entries on top
        // For some cases we want oldest entry on top (airgapped usage) where we pass 1
	filterKV.Value.SortOption = -1
}

lalat-das (changes requested):

pkg/apis/v1/api.proto
```diff
@@ -846,6 +846,15 @@ message TimeRange {
     google.protobuf.Timestamp end_time = 2;
 }
 
+// Sorting structure
+message SortOption {
+    // Name of the field to sort by
+    string field = 1;
+    // Sort order: ASCENDING or DESCENDING
+    int32 order = 2;
```
siva-portworx, line 854: @lalat-das I feel it is better to have enum as the normal use can understand it better and pass it.
Here, There is no information on what value that user need to pass it. If it enum, he will get the values automatically.
cc @prashanthpx

lalat-das: @siva-portworx  Yes I agree it should be enum, But I had commented earlier to make this enum such that the values directly mapped to the already defined mongo's sortOder values. Like -1 for ascending & 1 for descending. Anyone not filling it would make this variable set as zero.
And Zero indicates backend that we will follow default setting, and anything else would be assigned as it is.
Will alter this in the absence of @pallav-px .

PR #295, opened 2025-04-30
PB-10660 : Remove Identifiers field and add include_object and filters for inclusion and exclusion

vikasit12:


pkg/apis/v1/api.proto
```diff
@@ -1221,25 +1218,18 @@ message BackupScheduleDeleteRequest {
     // delete_backups is deprecated, not to be used.
     bool delete_backups = 3;
     string uid = 4;
-    //  ObjectRef can contain:
-    // - specific list of name+uid (e.g., 10 items)
-    // - wildcard: name = "*" or name = "pxb-*"
-    repeated ObjectRef identifiers = 5;
+    // specific list of name+uid (e.g., 10 items)
```
siva-portworx, line 1221: nit: Can you please maintain common comment pattern in all the places.

pkg/apis/v1/api.proto
```diff
@@ -1163,17 +1163,14 @@ message BackupScheduleUpdateRequest {
...
-
-    message BackupObjectType {
-        Type type = 1;
-        enum Type {
-            Invalid = 0;
-            All = 1; // for All application backup based on namespace
-            VirtualMachine = 2; // virtualMachine for VirtualMachine specific backup
-        }
-    }
+    // multiple name + uid for including
```
siva-portworx, line 1168: suggestion: Should we also say as "to pass static list of object to work on"

pkg/apis/v1/api.proto
```diff
@@ -1163,17 +1163,14 @@ message BackupScheduleUpdateRequest {
...
+    repeated ObjectRef include_objects = 24;
+    // Substring or regex pattern to match backup schedules and include (e.g. "*" for All, "pxb-" or any valid regex)
+    string include_filter = 25;
+    // Substring or regex pattern to match backup schedules and exclude (e.g. "*" for All, "pxb-" or any valid regex)
+    string exclude_filter = 26;
```
siva-portworx, line 1173: Should we also document what are the invalid combination, that we do not support? So it is clear in the proto file itself.

siva-portworx (approved):

pkg/apis/v1/api.proto
```diff
@@ -3091,3 +3081,21 @@ message ResourceTypeResponse {
...
+        // Represents backup schedules scoped to Kubernetes namespaces (non-VM workloads).
+        // Typically used for application-level backups.
+        NS = 1;
+
+        // Represents backup schedules scoped to VirtualMachine workloads.
+        // Used for VM-based backup scenarios.
+        VM = 2;
+
+        // Represents both NS and VM backup schedule types. Used when an operation should apply to all types.
+        All = 3;
```
vsundarraj-px, line 3097: We are changing the semantics of backupObjectType used elswhere. 
```message BackupObjectType {
        Type type = 1;
        enum Type {
            Invalid = 0;
            All = 1; // for All application backup based on namespace
            VirtualMachine = 2; // virtualMachine for VirtualMachine specific backup
        }
    }```
    If you dont give anything its All, if All is specified then its NS and VirtualMachine is 2. Can you keep it as is?

Comment (vikasit12) [reply pkg/apis/v1/api.proto:3097]: > We are changing the semantics of backupObjectType used elswhere.
> 
> ```
>         Type type = 1;
>         enum Type {
>             Invalid = 0;
>             All = 1; // for All application backup based on namespace
>             VirtualMachine = 2; // virtualMachine for VirtualMachine specific backup
>         }
>     }```
>     If you dont give anything its All, if All is specified then its NS and VirtualMachine is 2. Can you keep it as is? 
> ```

@vsundarraj-px if i set backupObjectType but not set the enum then by default it is Invalid as that is at 0 and as per my discussion with @siva-portworx and @prashanthpx i should never consider Invalid as All/Both, I agree I am deviating from the used value elsewhere but considering All Backups page I needed to provide All as 3rd option and put NS and VM separately for clearer representation.

Comment (vsundarraj-px) [reply pkg/apis/v1/api.proto:3097]: You are not introducing anything new. Currently also its the same thing right?
you will set backupInfo_backupObjectType to 1  for Namespace.
backupInfo_backupObjectType 2 for  is VM
if you do not set anything specifically you will not pass the filter, and that filter is not used at all. And hence both objects are selected.
The way to intepret is, 0 is not a default value but a invalid value. 
If you set 1 it means NS specifically
if you set 2 it means VM specifically
If you do not set anything, this field is not used for filter.. So all objects will be selected. 
Other way to interpret is, option is 1 always set, unless we specific 2

```backupObjectType := api.BackupScheduleInfo_BackupObjectType_All
	if req.GetBackupObjectType().GetType() == api.BackupScheduleCreateRequest_BackupObjectType_VirtualMachine {
		backupObjectType = api.BackupScheduleInfo_BackupObjectType_VirtualMachine
	}
```

vikasit12: > You are not introducing anything new. Currently also its the same thing right? you will set backupInfo_backupObjectType to 1 for Namespace. backupInfo_backupObjectType 2 for is VM if you do not set anything specifically you will not pass the filter, and that filter is not used at all. And hence both objects are selected. The way to intepret is, 0 is not a default value but a invalid value. If you set 1 it means NS specifically if you set 2 it means VM specifically If you do not set anything, this field is not used for filter.. So all objects will be selected. Other way to interpret is, option is 1 always set, unless we specific 2
> 
> ```
> backupObjectType := api.BackupScheduleInfo_BackupObjectType_All
> 	if req.GetBackupObjectType().GetType() == api.BackupScheduleCreateRequest_BackupObjectType_VirtualMachine {
> 		backupObjectType = api.BackupScheduleInfo_BackupObjectType_VirtualMachine
> 	}
> ```

@vsundarraj-px I agree to your above point but there is no clear demarkation in proto what is both, In your above statement we interpret but end user doesn't provide it specifically. For that reason I was introducing it and that to after @siva-portworx and @prashanthpx asked for it.

vsundarraj-px: Is this mandatory filed ? I think not in that are you planning to throw an error if this field is not set?
From API, lets keep with the assumption that when he intends to update, update is always for NS. Unless specifically specified as VM. Is that fair ? 
I am not seeing the need for new Type for this.

PR #297, opened 2025-05-02
PB-10547: Add flag for incremental backups

ss-px:
**What this PR does / why we need it**: Introduce `incremental_backup` field under volume info to indicate 
if the backup is full/incremental in case of PXD drivers.

Closes #PB-10547

**Special notes for your reviewer**: Need to set this field in a px-backup backend PR and test the changes.


pkg/apis/v1/api.proto
```diff
@@ -634,6 +634,8 @@ message BackupInfo {
         // As part of restrictive PSA setting, if an UID or GID is used during backup then same value should be used during restore (optional)
         JobSecurityContext job_security_context = 15;
         string virtual_machine_name = 16;
+        // Flag to indicate whether the backup is full or incremental in case of PXD driver
+        bool incremental_backup = 17;
```
kravichandran-px, line 638: IMO, we could change this to an enum and have 3 states: `unknown, full, incremental`. Though your code comment clearly conveys that this applies only for PXD driver, people using the API might misinterpret it to be  full/incremental based on what we set as default value. 
So converting to enum and setting unknown for non-pxd volumes might make it more explicit. 
@siva-portworx @prashanthpx Thoughts?

kravichandran-px (changes requested):

pkg/apis/v1/api.proto
vsundarraj-px: @kachaudhary-px I feel is_incremental should be the way to inform if a given backup is incremental or not. Unknown in my opinion is of no use to customer. He can not make any decision based on unknown. So we find a way to identify for a kinds of provisioner and get this field set accordingly or leave only with if its clearly marked incremental. I dont see any value in introducing unknown.

PR #300, opened 2025-05-05
PB-10547: Enumerate filter for backup mode

ss-px:
What this PR does / why we need it: Introduce backup_mode field under volume info to indicate
if the backup is full/incremental in case of PXD drivers or not supported in case of other drivers.

Closes #PB-10547

Special notes for your reviewer: Need to set this field in a px-backup backend PR and test the changes.

Older PR for reference: https://github.com/portworx/px-backup-api/pull/297


pkg/apis/v1/api.proto
```diff
@@ -634,11 +634,25 @@ message BackupInfo {
...
         string virtual_machine_name = 16;
+        // Indicate backup strategy - full/incr (for pxd volumes) or not supported (for other volume drivers)
+        BackupStrategy backup_strategy = 17;
 
         message JobSecurityContext {
             int64 runAsUser = 1;
             int64 runAsGroup = 2;
         }
+
+        message BackupStrategy {
```
vsundarraj-px, line 645: [Online Review comment] Please use BackupMode

pkg/apis/v1/api.proto
```diff
@@ -634,11 +634,25 @@ message BackupInfo {
...
+            Type type = 1;
+
+            enum Type {
+                // for non-pxd volumes
+                NotSupported = 0;
```
vsundarraj-px, line 650: default should always be Invalid. Use NotSupported as non 0 value.

vsundarraj-px (approved):
Please address review comments before merging

PR #301, opened 2025-05-05
Added support for sorting backup & restore entries

lalat-das:
**What this PR does / why we need it**: This PR enables sorting for backup & restore entries
    Fields used to sort by as mentioned below.
     - Name
     - ClusterName
     - Creation TimeStamp
     - Size
    The sorting order supported is ascending and descending.

Closes # PB-10525

even though keys for restore and backup objects are differing for size and cluster name, we are adding that intelligence in the backend code instead of expecting user to mention that  while filling API. Both approaches have pros and cons, we have chosen to follow common name.


pkg/apis/v1/api.proto
```diff
@@ -846,6 +846,26 @@ message TimeRange {
     google.protobuf.Timestamp end_time = 2;
 }
 
+message SortOption {
+    // Sorting field type
+    enum SortByType {
+        CreationTimestamp = 0;
```
siva-portworx, line 852: Zero should be invalid by default.

lalat-das: @siva-portworx  this particular case , I have made it that way to keep backward compatible. Today If someone doesn't send anything in enumerateOption people assume it as CreateTimestamp and Order as descending. I don't want to change that behavior in all the pages other that backup and restore.
I am saying other than backup restore page because in those pages our logic will go exclusively .

pkg/apis/v1/api.proto
```diff
@@ -846,6 +846,33 @@ message TimeRange {
...
+    // Order of sorting
+    message SortOrderType {
+        // Sorting order type
+        enum SortOrderType {
+            Invalid = 0;
+            Ascending = 1;
+            Descending = -1;
+        }
+    }
+    // Name of the field by which sort to be perfomed
```
siva-portworx, line 869: typo: perfomed

siva-portworx (approved):

PR #303, opened 2025-05-06
PB-10285 implement retry failed VM for failed and partial success VM backups

aks-px:
What this PR does / why we need it:
This PR exposes new gRPC call to retry failed VM backup
Closes #

Special notes for your reviewer:


pkg/apis/v1/api.proto
```diff
@@ -2262,6 +2271,29 @@ message BackupShareUpdateRequest {
     string uid = 4;
 }
 
+// Request message for retrying failed backup
+message BackupRetryRequest {
+    string org_id = 1;
```
dbinnal-px, line 2276: can we add comment string here also

dbinnal-px (approved):
LGTM

pkg/apis/v1/api.proto
```diff
@@ -2262,6 +2271,29 @@ message BackupShareUpdateRequest {
...
+    // UID of the failed or partial success backup to be retried
+    string uid = 3;
+    // List of resources to retry
+    // currently only supports VirtualMachine resources
+    repeated ResourceInfo include_resources = 4;
+    // Modify Skip auto rules for VirtualMachine Backup Object type
+    // This is only applicable for VirtualMachine backup object type
+    // Will update only if the original backup value differ. If original backup is set to true
+    // and if this field is set to true then no change is applied while retry. 
+    google.protobuf.BoolValue skip_vm_auto_exec_rules = 5;
```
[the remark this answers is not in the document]
siva-portworx: If the backup object is from schedulebackup, then the name itself will have a time stamp in it. Again, Are we planning to add new timestamp to it.
Now let us say, new retry had some couple of failed VM, If he retries how does the name will look like?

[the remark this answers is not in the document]
siva-portworx: Can a shared failed backup be retried?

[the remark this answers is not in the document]
vsundarraj-px: [review committee] review with PMO if this is option is required or not.

pkg/apis/v1/api.proto
```diff
@@ -2262,6 +2271,29 @@ message BackupShareUpdateRequest {
     string uid = 4;
 }
 
+// Request message for retrying failed backup
+message BackupRetryRequest {
+    string org_id = 1;
+    // Name of the failed or partial success backup to be retried
+    string name = 2;
```
vsundarraj-px, line 2278: [review committee] please add new field retry_parent in backupInfo to know this is retried object from the parent.

pkg/apis/v1/api.proto
```diff
@@ -2262,6 +2271,29 @@ message BackupShareUpdateRequest {
...
+    // UID of the failed or partial success backup to be retried
+    string uid = 3;
+    // List of resources to retry
+    // currently only supports VirtualMachine resources
```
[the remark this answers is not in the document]
vsundarraj-px: mention if this field is empty, we will retry all failed backups.

pkg/apis/v1/api.proto
```diff
@@ -2134,6 +2135,14 @@ service Backup {
             body : "*"
         };
     }
+    // RetryBackupResources retries backup of a failed or partial success backup
```
vsundarraj-px, line 2138: [review committee] add comment to specify what name will be created as part of retry

aks-px: Added in response message

pkg/apis/v1/api.proto
aks-px: Added

kgarg-px (approved):
LGTM

prashanthpx:
Putting my thoughts on the flow here
1. We're currently providing the option skip_vm_auto_exec_rules. This suggests that during a retry operation, users could deselect certain execution rules, which wasn't part of the original design discussion as I recall. If we proceed with this, as a user, I'd expect the UI to clearly indicate whether the backup is application-consistent or crash-consistent. Without this clarity, it's challenging for users to determine the backup consistency. Although it's technically possible to view the JSON and check the flag's status, from a user's perspective, this approach isn't user-friendly.

2. Additionally, for selecting all failed VMs, we currently imply that an empty include_resources means "select all failed VMs." This behavior is confusing, as typically, providing an empty include_resources in a backup workflow would signify that nothing should be backed up. I'm not comfortable with interpreting an empty value as "select all." Instead, we should either explicitly use a wildcard (*) or introduce a new variable to make the API usage clear.

prashanthpx (approved):

PR #309, opened 2025-05-12
PB-10547: Show PVC Backup incremental or full backup

ss-px:
What this PR does / why we need it: Introduce backup_mode field under volume info to indicate
if the backup is full/incremental in case of PXD drivers or not supported in case of other drivers.

Closes #PB-10547

Implemented by: https://github.com/pure-px/px-backup/pull/2118 (also contains testing details)


ss-px:
Merging with approval granted on: https://github.com/portworx/px-backup-api/pull/300

PR #326, opened 2025-06-09
PB-11106 Invalid param allowed during Backup/Schedule Object Create

vsundarraj-px:
This PR fixes bug PB-11196. IncludeResource uses resourceInfo structure and having virtualMachine name gives a ugly swagger input for POST calls
Closes #


pkg/apis/v1/api.proto
```diff
@@ -644,6 +648,8 @@ message BackupInfo {
     }
     // reference to volumeResourceOnly policy object
     ObjectRef volume_resource_only_policy_ref = 51;
+    // resourcesInfo for backupObjectType VirtualMachine
+    repeated VirtualMachineResourceInfo virtual_machine_resources = 52;
```
siva-portworx, line 652: Can you please add a comment for below field that it will be used for backupObjectType Namespace. 
 repeated ResourceInfo resources = 6;

pkg/apis/v1/api.proto
```diff
@@ -553,9 +553,13 @@ message ResourceInfo {
     string group = 3;
     string kind = 4;
     string version = 5;
-    repeated string virtual_machine_name = 6;
 }
 
+// ResourcesInfo used for VM backups
+message VirtualMachineResourceInfo {
+    ResourceInfo resource_info = 1 [ (gogoproto.embed) = true ];
+    repeated string virtual_machine_name = 2;
```
siva-portworx, line 561: nit: virtual_machine_name -> virtual_machine_names

siva-portworx: Here it is an array because same resource can be used by multiple VMs?

siva-portworx (approved):

t-nidhin (approved):

pkg/apis/v1/api.proto
vsundarraj-px: Thats correct

PR #330, opened 2025-07-21
PB-11569 | Add status info to ResourceInfo in API definition

mkoppal-px:

Closes #


pkg/apis/v1/api.proto
```diff
@@ -555,6 +555,27 @@ message ResourceInfo {
...
     string version = 5;
+    StatusInfo status = 6;
+    
+    message StatusInfo {
+        Status status = 1;
+        string reason = 2;
+
+        enum Status {
+            Invalid = 0;
+            Pending = 1;
```
prashanthpx, line 566: Do we need such extensive status? As we will have only two status "Success" or "Failed"

PR #334, opened 2025-08-07
PB-11835 : Add remarks field to BackupScheduleUpdateRequest and define Remark message

vikasit12:
- Introduced a new field `remark` in `BackupScheduleUpdateRequest` and `BackupScheduleInfo` to capture user comments during update operations.
- Defined a new `Remark` message to structure the remark data, including fields for the comment text, username of the creator, and timestamp.

mechanism to add a note for schedules which will be suspended / resumed.

One action will have one input field for the user ie, if a user selects bunch of schedules for suspension, there will be just one input box for him to enter the reason for suspension / resuming.

There should be a place where they can see why schedules were disabled.


pkg/apis/v1/api.proto
```diff
@@ -2424,6 +2428,16 @@ message BackupUpdateRequest {
...
 message BackupUpdateResponse {}
 
+// Remark represents a user comment why is he performing the action
+message Remark {
+    // The actual remark/comment text
+    string comment = 1;
+    // Username of the person making the remark
+    string updated_by = 2;
+    // Timestamp when the remark was made
+    google.protobuf.Timestamp updated_at = 3;
```
pallav-px, line 2438: updated_at and updated_by should be part of Metadata can be use later in every object

pkg/apis/v1/api.proto
```diff
@@ -1393,6 +1395,8 @@ message BackupScheduleUpdateRequest {
     ClusterScope cluster_scope = 27; 
     // reference to volumeResourceOnly policy object
     ObjectRef volume_resource_only_policy_ref = 28;
+    // User remarks for this update operation
+    string remark = 29;
```
t-nidhin, line 1399: why not `Remark` here, but using a string?

vikasit12: > why not `Remark` here, but using a string?

@t-nidhin 
I am taking input just a comment from user rather than his name and time but in response i am setting who updated the comment and when hence here its just a string but in `BackupScheduleInfo` it is Remark rather than string as in response structure is filled

aks-px (approved):

pkg/apis/v1/api.proto
vikasit12: > IMO, we should either change this to:
> 
> 1. `string comment` //aligns with the param of Remark.comment. Let's say you add something in the future, then `string remark = 29` alone becomes meaningless if we need to pass another param inside remark
> 2. Try finding a different naming convention `remarkComment` or `remark_comment`
> 3. Add more explanation in comment; what you expect the user to pass here, mention the reference to `message Remark.comment`. Since sometime we require users to directly refer this. Helps in case of Swagger.

remark itself means to say or write something; to comment
It shouldn't be meaningless

px-kesavan (approved):

PR #335, opened 2025-08-10
PB-11619: api changes for adding cluer sort options and create time

dbinnal-px:
PB-11619: This PR is to support cluster sorting and cluster create time from backend.


pkg/apis/v1/api.proto
```diff
@@ -80,6 +80,8 @@ message ClusterInfo {
     bool delete_all_cluster_backups = 22;
     // time_zone is the time zone of the cluster
     string time_zone = 23;
+    // create_time is the time when the cluster was created
+    google.protobuf.Timestamp create_time = 24;
```
aks-px, line 84: This field is already in Metadata,hence not required to be added

PR #341, opened 2025-08-31
Mege 2.10.0-staging proto changes to master

shkumari-px:

Closes #


pkg/apis/v1/api.proto
```diff
@@ -568,7 +572,7 @@ message BackupInfo {
     string backup_location = 1;
     // Name of Cluster
     string cluster = 2;
-    // Namespaces to backup. Only an admin can provide multiple namespaces
+    // Namespaces to backup. Only an admin can provide multiple namespaces, this will be updated to the namespaces present at the start of backup cycle
```
mkoppal-px, line 575: nit: this will be updated `with` the namespaces present at the start of the backup cycle

PR #345, opened 2025-09-10
PB-12247 Define List VM backups gRPC call with request and response s…

vsundarraj-px:
This PR implements PB-12247 where VM and or Resource Filter can be passed to BackupEnumerate gRPC call which will filter the backups that have the specified resource backed up. 
Closes #
PB-12247


pkg/apis/v1/api.proto
```diff
@@ -999,6 +999,12 @@ message EnumerateOptions {
     // If the field is not set, default value will be "CreationTimestamp"
     // If the order is not set, default value will be "descending"
     SortOption sort_option = 14;
+    // filter to use resource name and namespace. Any backup that contains the resource will be returned
+    ResourceInfo resource_info = 15;
+    // filter VM that matches the resource_info and has volume vm_volume_name attached to it
+    string vm_volume_name = 16;
```
kravichandran-px, line 1005: Could you please share the usecase to search by volume name for VM? From what I understand, the use case is to list volumes of a backed-up VM during sfr workflow, but not for VM vol name based backup enumeration. Please correct me if Im wrong.

pkg/apis/v1/api.proto
```diff
@@ -999,6 +999,12 @@ message EnumerateOptions {
...
+    // filter to exclude failed backup resources while enumerating backups
+    bool exclude_failed_backup_resource = 17;
```
kravichandran-px, line 1007: Only this field has "backup" in it while other fields do not :) Could we make it generic, unless intentional?

PR #348, opened 2025-09-27
PB-12297 | Proto changes for resource split feature

mkoppal-px:
**This PR contains the following changes:**
- Deprecate repeated `VirtualMachineResourceInfo` `virtual_machine_resources` field
- Add oneof `resource_representation` with `VirtualMachineResources` and `NamespaceResources` options
- Introduce `NamespaceResource` structure with hierarchical resource organization
- Add `NamespaceStatusInfo` enum with Invalid, InProgress, Success, Failed, PartialSuccess
- Add `ChunkStatusInfo` enum with Invalid, Success, Failed
- Use int32 for `chunk_id` to support sequential numbering (0,1,2,3...)

This PR is needed to accommodate chunk info as part of resources and maintain an optimised structure for resources in BackupInfo with minimal duplicate fields but also at the same time to make mongo query simpler to help with filtering in case of granular restore feature. 

Closes #PB-12297

This is a critical proto change which will shape the logic in the way stork captures and send the resource details to Px-Backup through ABCR and also critical in making a decision of which fields in mongodb to migrate.


pkg/apis/v1/api.proto
```diff
@@ -656,9 +656,68 @@ message BackupInfo {
...
+
+    // Message for Namespace resources
+    message NamespaceResources {
+        repeated NamespaceResource namespace_resources = 1;
+    }
+
+    // Message representing resources within a namespace
+    message NamespaceResource {
+        string namespace = 1;
+        NamespaceStatusInfo namespace_status = 2;
```
siva-portworx, line 683: Does namespace status is determined by including volume and resource completion status? or only the resource status?
If it is only resource level, it might be confusing to the user that volume level it is partial success and resources level it is success.

pkg/apis/v1/api.proto
```diff
@@ -656,9 +656,68 @@ message BackupInfo {
...
-    repeated VirtualMachineResourceInfo virtual_machine_resources = 52;
+    // DEPRECATED: Use resource_representation instead
+    repeated VirtualMachineResourceInfo virtual_machine_resources = 52 [deprecated = true];
    // to capture namespaces that were missing when the backup was initiated
     repeated string missing_namespaces = 53;
+
+    // New resource representation using oneof for different resource types
+    oneof resource_representation {
+        VirtualMachineResources vm_resources = 54;
+        NamespaceResources namespace_resources = 55;
```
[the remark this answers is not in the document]
siva-portworx: Will the NamespaceResources will be empty for the large resource and Restore granularity will use this structure to sync the resources?

lalat-das:
@mkoppal-px Why this commit is against 2.10-staging. Isn't it that we are maintaining feature branch for resource split in every repo we are touching... I am curious because we need to raise PR for granular restore which should follow the same feature branch of resource split. Am I missing sth

pkg/apis/v1/api.proto
siva-portworx: For PVC as a resourceType, the status will be successful instead of partial success.
However, since volume backup failed for some of the volumes, the namespace status should be shown as partialSuccess.

ChunkInfo → will still show full success.

Namespace status → should show partialSuccess (because of volume-level failures).

It may be better to introduce a flag indicating that a volume-level backup failure occurred, which caused the namespace status to be reported as partial success.

pkg/apis/v1/api.proto
```diff
@@ -656,9 +656,68 @@ message BackupInfo {
...
+        }
+        Status status = 1;
+        string reason = 2;
+    }
+
+    // Status information for chunk-level operations
+    message ChunkStatusInfo {
+        enum Status {
+            Invalid = 0;
+            Success = 1;
```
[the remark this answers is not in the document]
mkoppal-px: Added InProgress, but I do not think we will be using it. We can take a call to remove it at the end of code complete in case we decide not to go with it

aks-px (approved):

pkg/apis/v1/api.proto
```diff
@@ -656,9 +656,74 @@ message BackupInfo {
...
+    // Message representing resources within a namespace
+    message NamespaceResource {
+        string namespace = 1;
+        NamespaceStatusInfo namespace_status = 2;
+        repeated ResourceTypeInfo resource_types = 3;
+
+        // ResourceTypeInfo groups resources by their Kubernetes resource type (e.g., "Pod", "Service", "ConfigMap")
+        // Each resource type contains multiple chunks for processing efficiency
+        message ResourceTypeInfo {
+            string resource_type = 1;
```
aks-px, line 689: Will this be like g/v/k or just kind?

pkg/apis/v1/api.proto
```diff
@@ -656,9 +656,74 @@ message BackupInfo {
...
+    }
+
+    // Message for VirtualMachine resources
+    message VirtualMachineResources {
+        repeated VirtualMachineResourceInfo virtual_machine_resources = 1;
+    }
+
+    // Message for Namespace resources
+    message NamespaceResources {
+        repeated NamespaceResource namespace_resources = 1;
```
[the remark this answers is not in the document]
mkoppal-px: Fixed

pkg/apis/v1/api.proto
mkoppal-px: Improved the code comment

pkg/apis/v1/api.proto
```diff
@@ -656,9 +657,98 @@ message BackupInfo {
...
+
+    // New resource representation using oneof for different resource types
+    oneof resource_representation {
+        VirtualMachineResources vm_resources = 54;
+        NamespaceResources namespace_resources = 55;
+    }
+
+    // Message for VirtualMachine resources
+    message VirtualMachineResources {
+        repeated VirtualMachineResourceInfo virtual_machine_resources = 1;
```
px-kesavan, line 673: i would suggest to keep the variable name as `resources` here as well. having structure called virtual_machine_resources under vm_resources feels redundant. Anyway oneof cannot have repeated type , so something like below will suffice 

```
vm_resources {
    resources {
    .
    .
    
namespace_resources {
    resources {
    .
    .
    .

```

pkg/apis/v1/api.proto
```diff
@@ -656,9 +657,98 @@ message BackupInfo {
...
+    message NamespaceResources {
+        repeated NamespaceResource resources = 1;
+    }
+
+    // Message representing resources within a namespace
+    message NamespaceResource {
+        string namespace = 1;
+        NamespaceStatusInfo namespace_status = 2;
+        repeated ResourceTypeInfo resource_types = 3;
+        NamespaceStats namespace_stats = 4;
```
px-kesavan, line 686: nit:

got confused with namespace_status and namespace_stats , may be a better name would differentiate better

PR #400, opened 2026-03-07
PB-14767 : Add force option to BackupDeleteRequest for orphaned metadata cleanup

vikasit12:
- Introduced a `force` boolean field in the BackupDeleteRequest message to allow the removal of backup metadata when the underlying cluster is no longer available.
- Updated the API Swagger documentation to include the new `force` parameter, detailing its usage and implications for data deletion.

Last resort to move the backup shown in UI out (specifically in case of managed clusters)

Closes #PB-14767

Details why we need and applied this is captured at https://purestorage.slack.com/archives/C09S8HTRPN1/p1771849623075449


pkg/apis/v1/api.proto
```diff
@@ -3000,6 +3000,12 @@ message BackupDeleteRequest {
     string cluster = 3;
     string uid = 4;
     ObjectRef cluster_ref = 5;
+    // force allows the removal of backup metadata when the underlying cluster is gone.
+    // WARNING: This only deletes the record from MongoDB and skips physical cleanup
+    // of the backup location. To ensure full data deletion, prefer registering a
+    // temporary cluster with the same backup location. Use this flag only for
+    // orphaned metadata where the actual backup data is no longer needed.
+    bool force = 6;
```
siva-portworx, line 3011: Since this force flag is going to be used between federated and non-federated, It would be good, if you can explain the behavior across both the federated and non-federated separately.

vikasit12: > Since this force flag is going to be used between federated and non-federated, It would be good, if you can explain the behavior across both the federated and non-federated separately.

@siva-portworx I am adding below in the PR

    // NOTE: This option is only valid in federated mode and will return an error
    // in non-federated mode.

mkoppal-px (approved):

PR #403, opened 2026-03-11
feat(api): Add bidirectional BackupLocation-Cluster references for WLI mode

pallav-px:
- Rename federated to use_workload_identity in BackupLocationInfo
- Add cluster_refs and cluster_status for per-cluster validation tracking
- Add BackupLocationRefStatus with ObjectRef+StatusInfo to ClusterInfo for reverse lookup
- Add enumerate filters: cluster_ref on BL, backup_location_ref on Cluster

Closes #

### DIFF
--- a/pkg/apis/v1/api.proto
+++ b/pkg/apis/v1/api.proto
@@ -85,6 +85,11 @@ message ClusterInfo {
     // by external providers (e.g., Gardener).
     // This field is populated based on how the cluster was discovered/added.
     ClusterProviderInfo provider_info = 24;
+    // List of BackupLocations that reference this cluster with their validation status.
+    // This enables bidirectional reference: Cluster → BackupLocation
+    // System-managed field (not user input) - populated when BackupLocations are created/updated
+    // with this cluster in their cluster_refs list.
+    repeated BackupLocationRefStatus backup_location_ref_status = 25;
 
     // ClusterProviderInfo is a polymorphic wrapper for provider-specific cluster information.
     // Use the oneof to determine which provider type is set.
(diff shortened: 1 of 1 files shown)


sodey-px (approved):

pkg/apis/v1/api.proto
```diff
@@ -640,14 +645,21 @@ message completionTimeInfo {
     google.protobuf.Timestamp total_completion_time = 3;
 }
 
+// ClusterValidationStatus is the value in BackupLocationInfo.cluster_status map
+// Key is "name-uid" or "name", so no need for cluster_name field
```
mkoppal-px, line 649: Having name-uid, will it not cause problem if you want to split the name and UID using the key?

pallav-px: uid is not a mandatory field and name is not unique. This status will just hold the status of the cluster. For enumerate filter we are going to use `repeated cluster_ref `in spec field

pkg/apis/v1/api.proto
```diff
@@ -2126,6 +2178,11 @@ message ClusterEnumerateRequest {
     bool only_backup_share = 6;
     // Optional arguments for enumeration
     EnumerateOptions enumerate_options = 7;
+    // Optional filter to return only clusters that are referenced by the specified BackupLocation.
+    // If not provided, return ALL clusters (existing behavior).
+    // If provided, return ONLY clusters that have this BackupLocation in their backup_location_ref_status.
+    // This enables bidirectional querying: "Show me all clusters that can access this BackupLocation"
+    ObjectRef backup_location_ref = 8;
```
siva-portworx, line 2185: @mkoppal-px how are going to take care of Vijay usecase of find the cluster list which does not have any backuplocationRef association?

mkoppal-px (approved):

PR #407, opened 2026-03-18
PB-14925 | Update ClusterDiscoveryConfigType enum and add `DeletePending` state in API proto

mkoppal-px:
---

#### 1. Rename `ClusterDiscoveryConfigTypeInvalid` → `ClusterDiscoveryConfigTypeUnspecified`
- Renamed the zero-value enum entry to follow standard Protobuf conventions
- `Unspecified` better conveys the semantics: when no type is set in enumerate/filter requests, all discovery config types are returned
- Removed the "Future types" comment from the enum as it's unnecessary

#### 2. Add `DeletePending` state to `ClusterDiscoveryConfigInfo.StatusInfo`
- Added `DeletePending = 5` to the `Status` enum
- This state represents a discovery config that has been marked for deletion but hasn't been fully cleaned up yet

### Files Changed
- `pkg/apis/v1/api.proto` — Enum rename and new status value
- `pkg/apis/v1/api.pb.go` — Regenerated Go code
- `pkg/apis/v1/api.swagger.json` — Updated Swagger/OpenAPI spec

### Related Jira
PB-14925

---


pallav-px (approved):

vikasit12 (approved):
@mkoppal-px this totally makes sense but this is for sure a conflict zone.

As per Px-Backup long lived rituals 0th value has to be Invalid always (which I am not a firm believer of)

mkoppal-px:
> @mkoppal-px this totally makes sense but this is for sure a conflict zone.
> 
> As per Px-Backup long lived rituals 0th value has to be Invalid always (which I am not a firm believer of)

@vikasit12 I got a comment from @siva-portworx here - https://github.com/pure-px/px-backup/pull/2594#discussion_r2940818087
We cannot use invalid as a type. But using "All" as a type for filtering also does not make sense because it is not standard protobuf practice. Also, having "All" as a separate type for an interface is a bad use of interface. 

Everywhere in our proto I see patterns like this for oneof

```
    oneof Config {
        AWSConfig aws_config = 100;
        GoogleConfig google_config = 101;
        AzureConfig azure_config = 102;
    }
```

There is no "All" type separately. @siva-portworx. Please advise on how to go about this.
Also I need this PR for my delete API as I need the "Delete Pending" state.

mkoppal-px:
@siva-portworx You wanted me to do it like this?

```
message BackupObjectType {
    enum Type {
        Invalid = 0;

        // Represents backup schedules scoped to Kubernetes namespaces (non-VM workloads).
        // Typically used for application-level backups.
        NS = 1;

        // Represents backup schedules scoped to VirtualMachine workloads.
        // Used for VM-based backup scenarios.
        VM = 2;

        // Represents both NS and VM backup schedule types. Used when an operation should apply to all types.
        All = 3;
    }

    Type type = 1;
}
```

yassharma-px (approved):

siva-portworx (approved):

PR #412, opened 2026-03-26
Add service for refreshing shoot kubeconfigs

yassharma-px:
## What this PR does / why we need it:

Adds a RefreshClusters RPC to the ClusterDiscoveryConfig service, enabling on-demand credential refresh for all clusters associated with a given discovery configuration.

Gardener Shoot cluster kubeconfigs have a 24-hour TTL and currently only auto-refresh at the 23rd hour. This API provides a manual trigger for cases where credentials are revoked or need immediate rotation, without waiting for the automatic refresh cycle.

## Changes:
- New RefreshClusters RPC in the ClusterDiscoveryConfig service
- REST endpoint: POST /v1/clusterdiscoveryconfig/{org_id}/{name}/refresh
- New request/response messages: ClusterDiscoveryConfigRefreshClustersRequest / ClusterDiscoveryConfigRefreshClustersResponse
- Regenerated Go code and Swagger definitions

Special notes for your reviewer:
- The RPC is named generically (RefreshClusters) rather than RefreshKubeconfigs since ClusterDiscoveryConfig may support non-Shoot discovery types in the future.
- Follows the same async trigger pattern as the existing DiscoverClusters RPC.


adasingh-px (approved):

pkg/apis/v1/api.proto
```diff
@@ -4444,6 +4444,19 @@ service ClusterDiscoveryConfig {
             body : "*"
         };
     }
+
+    // RefreshClusters manually triggers an on-demand credential refresh for all
+    // clusters associated with the specified ClusterDiscoveryConfig.
+    // For Gardener Shoot clusters, this refreshes the kubeconfig for each cluster.
+    // The refresh runs asynchronously; use Cluster Inspect to check updated
+    // credential timestamps (e.g., expiration_timestamp, kubeconfig_obtained_at).
```
mkoppal-px, line 4453: I think we should mention that we support this only for gardener shoot for now

mkoppal-px (approved):

pkg/apis/v1/api.proto
yassharma-px: updated

PR #413, opened 2026-03-26
PB-14587 | Add `managed` field to GardenerShootInfo and update `auto_discover` to use BoolValue in API proto

mkoppal-px:
#### 1. Add `managed` field to `GardenerShootInfo`
- Added `bool managed = 9` to `ClusterInfo.GardenerShootInfo`
- This field tracks whether a Gardener Shoot cluster's required onboarding label still exists on the Shoot CR
  - `true`: The required label exists and the cluster is actively managed by PX-Backup
  - `false`: The cluster is managed by PX-Backup but the required label has been removed externally (e.g., by a user or automation)

#### 2. Change `auto_discover` to `google.protobuf.BoolValue` in `ClusterDiscoverySettings`
- Changed `bool auto_discover` to `google.protobuf.BoolValue auto_discover`
- Added `google/protobuf/wrappers.proto` import
- **Reason:** With a plain `bool`, there is no way to distinguish between "user didn't provide the field" and "user explicitly set it to false." This caused an issue during update operations where omitting `auto_discover` would reset it to `false`, unintentionally disabling auto-discovery. Using `BoolValue`, the update logic can now:
  - `nil` → field not provided, retain existing value
  - `true` → user explicitly enabled auto-discovery
  - `false` → user explicitly disabled auto-discovery

### Related Jira
PB-14587


px-kesavan (approved):

pkg/apis/v1/api.proto
```diff
@@ -176,6 +177,11 @@ message ClusterInfo {
         // Timestamp when the kubeconfig was obtained from Gardener.
         // Used to track kubeconfig freshness and determine refresh timing.
         google.protobuf.Timestamp kubeconfig_obtained_at = 8;
+        // Indicates whether the Gardener Shoot cluster is actively managed by PX-Backup.
+        // Set to true if the required label used to onboard this cluster still exists
+        // on the Shoot CR. Set to false if the cluster is managed by PX-Backup but
+        // the required label is missing (e.g., removed externally).
```
siva-portworx, line 183: nit:  the required label is missing on the shoot CR.

one question. Will this cluster with managed equal to false will be included for cluster limit caculcation for a given instance of PXB?

siva-portworx (approved):

pkg/apis/v1/api.proto
```diff
@@ -176,6 +177,11 @@ message ClusterInfo {
...
+        bool managed = 9;
```
adasingh-px, line 184: Shall we also add this in `ClusterEnumerateOptions` for filtering during Cluster Enumerate?

pallav-px (approved):

pkg/apis/v1/api.proto
mkoppal-px: No Siva. Cluster limit calculation is basically enumerating all the clusters across Px-Backup (irrespective of whether it is online/offline, managed/unmanaged, gardener/vanilla/ocp/aks etc.,)

pkg/apis/v1/api.proto
mkoppal-px: Yes, will do

adasingh-px (approved):

PR #414, opened 2026-03-28
PB- 15040 Update API switched to PATCH from PUT for clusterdiscoveryconfig

adasingh-px:
### API Proto (`api.proto`)
- Changed HTTP method from `put` to `patch`
- Added `{metadata.org_id}/{metadata.name}` path parameters to the URL

## PATCH Semantics

- **Updatable fields:** `settings.auto_discover`, `settings.auto_discover_frequency`, `metadata.labels`, `shoot_config.gardener_kubeconfig` (token/creds only)
- **Non-updatable fields (rejected):** `project_name`, `label_selector`, gardener API server endpoint
- **Required:** `confirm_update: true`

https://github.com/pure-px/px-backup/pull/2625


pkg/apis/v1/api.proto
```diff
@@ -4481,7 +4481,11 @@ message ClusterDiscoveryConfigCreateResponse {
 }
 
 message ClusterDiscoveryConfigUpdateRequest {
-    CreateMetadata metadata = 1 [ (gogoproto.embed) = true ];
+    // Metadata for the cluster discovery config to update.
+    // name and org_id (from URL path) are required. uid is optional for exact match.
+    // labels are updatable. Server-managed fields are ignored.
+    Metadata metadata = 1;
```
mkoppal-px: Sorry, I forgot to ask you this. If I do not want to change the labels, even then I need to provide the name and org both in the URL and request body?

mkoppal-px (approved):

yassharma-px (approved):

siva-portworx (approved):

PR #415, opened 2026-03-31
PB-15019 add fields for tracking manual refresh status

yassharma-px:
Adds fields for tracking manual refresh status to allow both discovery and refresh operations to run simultaneously.


pkg/apis/v1/api.proto
```diff
@@ -4392,6 +4396,19 @@ message ClusterDiscoveryConfigInfo {
             DeletePending = 5;
         }
     }
+
+    // Message for maintaining status of the manual refresh operation.
+    message RefreshStatusInfo {
+        RefreshStatus status = 1;
+        string reason = 2;
```
mkoppal-px, line 4403: Can we also add two more stats to this? `processed` and `total`?
We can see the progress like `23 out of 75` done

mkoppal-px (commented):
Reviewed

pkg/apis/v1/api.proto
yassharma-px: updated

mkoppal-px (approved):

PR #428, opened 2026-06-24
PB-14450: Add VM related status info per VM for partial success in case of VM related restores

kgarg-px:
Add VM related status info per VM for partial success in case of VM related restores
Closes #
PB-14450


pkg/apis/v1/api.proto
```diff
@@ -1366,6 +1366,14 @@ message RestoreInfo {
         bool skip_vm_restart = 2;
     }
 
+    // Per-VM restore outcome for VirtualMachine-typed restores.
+    // status carries both the per-VM Status enum (including PartialSuccess)
```
siva-portworx: Second line comment is not clear. Why are we referring  to PartialSuccess? Per VM status will not have Partial Success status rt?

pkg/apis/v1/api.proto
```diff
@@ -1379,6 +1387,10 @@ message RestoreInfo {
     // retention_period stores the object-lock retention period in days for the restore
     // This is only set when the backup location has object lock enabled
     int64 retention_period = 32;
+    // Per-VM restore outcome for VirtualMachine-typed restores.
+    // Empty for namespace-mode restores or pre-feature restores; new clients
+    // should treat absence as "no per-VM data available".
+    repeated VirtualMachineRestoreInfo virtual_machines = 33;
```
siva-portworx: Nit: Would suggest to name the variable as virtual_machines_status to indicate it is not just of vm, It is the status of restore vms

pkg/apis/v1/api.proto
```diff
@@ -1366,6 +1366,14 @@ message RestoreInfo {
         bool skip_vm_restart = 2;
     }
 
+    // Per-VM restore outcome for VirtualMachine-typed restores.
+    // status carries both the per-VM Status enum (including PartialSuccess)
+    // and the reason string, mirroring BackupInfo.VirtualMachineInfo.
+    message VirtualMachineRestoreInfo {
+        string name = 1;
+        string namespace = 2;
+        StatusInfo status = 3;
```
siva-portworx: Just thinking, To save space, should we consider different format to avoid the namespace and status is repeated for every VM entry. But that will make namespace as primary key, Which might be confusing.
cc: @vsundarraj-px

siva-portworx (approved):

pkg/apis/v1/api.proto
vsundarraj-px: Yes we should look at optimising this.
status is required. This is moving value between stages. So we cannot have vms under status field as they transition form pending->inProgress->failed/success

However we should leverage reason_index logic currently exist from resource split implementation. Many VMs can fail with same reason. So indexing the reason will optimise here.

vsundarraj-px: 2. We should also index namespace. Instead of storing namespace in each of the VM object. This will optimise a lot. 

3. To accommodate this, we will need to add array VirtualMachineIndex in Resource structure and VolumeInfo structure to point to which VM these resources are referenced.

pkg/apis/v1/api.proto
```diff
@@ -3380,6 +3473,90 @@ message BackupResourceObject {
...
+    repeated RestoreInfo.RestoredResource resources = 17;
+    // Total count of resources matching the filter (for pagination).
+    uint64 total_resource_count = 18;
+    // Namespaces filtered for this restore.
+    repeated FilteredNamespaceInfo filtered_namespaces = 19;
+
+    // Embedded dedup tables — clients resolve VolumeOutcome.namespace_index,
+    // ResourceOutcome.resource_type_index, VolumeOutcome.reason_index, etc.,
+    // via single map lookup. Index 0 is the sentinel empty string in each.
+    repeated string namespaces     = 21;
```
vsundarraj-px: These should not be part of RestoreResourceObject() this object is not persistent data but built data on runtime from restoreInfo. The index optimisation should only be part of restoreInfo not in this.

pkg/apis/v1/api.proto
```diff
@@ -3380,6 +3473,90 @@ message BackupResourceObject {
...
+
+    // VM-mode counters + aggregates (mirror the additions on RestoreInfo).
+    int64 virtual_machines_total_count = 12;
+    int64 virtual_machines_failed_count = 13;
+    // VM-centric: one entry per VM, each with its embedded RestoreOutcome
+    // (counts + Failed/Skipped/Retained entries; Successful entries gated by
+    // the request's include_successful flag).
+    repeated RestoreInfo.VirtualMachineRestoreInfo virtual_machines = 14;
+
+    // Filtered/paginated detail arrays.
```
vsundarraj-px, line 3510: These two are redundant needed for backward compatibility only for restore objects that were taken on older stork. Lets have the comment say the same.

pkg/apis/v1/api.proto
kgarg-px: addressed

pkg/apis/v1/api.proto
kgarg-px: addressed

pkg/apis/v1/api.proto
kgarg-px: the structure of proto is different now, so resolving this comment

pkg/apis/v1/api.proto
kgarg-px: the structure of proto is different now, so resolving this comment

pkg/apis/v1/api.proto
```diff
@@ -1345,6 +1398,11 @@ message RestoreInfo {
             Success = 6;
             Retained = 7;
             PartialSuccess = 8;
+            Skipped = 9;
+            // Deleted applies to a previously-Successful per-volume / per-resource
+            // entry that stork's cleanup sub-phase removed after its owner VM
+            // ended Failed.
```
siva-portworx: This comment is too internal to stork flow, which normal user might not understand. Better to make it generic. 

```
// Deleted indicates a successfully restored volume or resource
// entry that was later removed during cleanup because its owning
// object ultimately failed.
```

pkg/apis/v1/api.proto
```diff
@@ -878,6 +878,59 @@ message VirtualMachineResourceInfo {
...
+// resources are unified into one list, differentiated by gvk.
+// For PVCs, gvk = "PersistentVolumeClaim/v1".
+message ResourceDetail {
+    string                 name   = 1;
+    string                 gvk    = 2;  // "Kind/Group/Version"; empty group = core
+    RestoreInfo.StatusInfo status = 3;  // status enum + resolved reason string
+}
+
+// VolumeStats and ResourceStats are per-VM count breakdowns by terminal status.
+// All values are computed at response build time by px-backup from the CR's
```
siva-portworx: It is better not give internal details in the proto definition. As a generic user, he might be aware of the internal of the px-backup and CR.

pkg/apis/v1/api.proto
```diff
@@ -878,6 +878,59 @@ message VirtualMachineResourceInfo {
...
+// flat entry arrays; never persisted.
+message VolumeStats {
+    int32 total    = 1;
+    int32 failed   = 2;
+    int32 skipped  = 3;
+    int32 retained = 4;
+    int32 deleted  = 5;
+}
+
+message ResourceStats {
```
siva-portworx: Since stats is same for Volume and Resource, we can make it as generic Stats message structure and use for the volume and resource definition.

pkg/apis/v1/api.proto
```diff
@@ -3380,6 +3476,88 @@ message BackupResourceObject {
...
+    ObjectRef backup_ref = 10;
+    BackupObjectType backup_object_type = 11;
+
+    // Aggregate VM-mode counters retained for backward compatibility with
+    // restores taken on older stork (pre-partial-success), where the per-VM
+    // virtual_machines[] array below is absent — these counters still allow
+    // consumers to surface total/failed numbers without iterating the array.
+    // For new restores, these are derivable from virtual_machines[] but kept
+    // for self-contained response semantics.
+    int64 virtual_machines_total_count  = 12;
```
vsundarraj-px, line 3504: Nit: already there is stats part of existing restoreInfo under resources.Metrics We should the same. The metrics are at resource level only does not capture at high level. So We can one inside restoreInfo.Metrics With Total Volumes, Total Resources, Total VirtualMachines, Failed Volumes, Failed Resources, Failed VirtualMachines.

vsundarraj-px (approved):
LGTM. With some comments that I would prefer to take it up outside of spec kit. 
May be as improvement task which will not need spec generated code.

pkg/apis/v1/api.proto
```diff
@@ -1366,6 +1424,18 @@ message RestoreInfo {
...
+    // Per-VM persistent shape mirrored from stork CR.
+    // Returned by Restore.Inspect; matches mongo storage shape.
+    // reason_index is a 0-based position into RestoreInfo.reasons[].
+    // vol_index / resource_index are 0-based positions into Status.Volumes[] /
+    // Status.Resources[] (append-only — those arrays must never reorder).
+    message VirtualMachineStatus {
+        VirtualMachine    vm             = 1;
+        StatusInfo.Status status         = 2;
+        int32             reason_index   = 3;
+        repeated int32    vol_index      = 4;
```
siva-portworx: Not sure whether the index is really going to be useful, when we decided to give GetRestoreResourceDetails API. Will discuss tomorrow, when I am in office.

pkg/apis/v1/api.proto
kgarg-px: addressed

pkg/apis/v1/api.proto
kgarg-px: addressed

pkg/apis/v1/api.proto
kgarg-px: Indices were added based on the discussion that I had with @vsundarraj-px on friday and monday.
Below are the reasons why it was decided to keep (Diverging or going back would also mean letting go of 3 day effort).
Reason is as below:
The indices serve a purpose distinct from the GetRestoreResourceDetails API. Two reasons to keep them:

1. Indices live on the persistent / Inspect layer, not on GetRestoreResourceDetails response.

The response from GetRestoreResourceDetails (RestoreResourceObject → VirtualMachineDetail) carries fully resolved strings — no indices visible. The indices live on RestoreInfo and RestoreInfo.VirtualMachineStatus, which is what:

Mongo stores per-restore document
Restore.Inspect returns (the existing, frequently-called inspection RPC)
Dropping indices means every Inspect response and every mongo document carries fully resolved strings, which has real cost at scale.

2. reason_index is a dedup win, not a presentation choice.

Reasons can be up to 1 KB each (PB-16170 cap). At 1000-VM × shared-cause failure ("BackupLocation unreachable"), the dedup is:

Without index: 1000 × 1 KB = 1 MB of repeated reason text in one document
With index: 1 KB (the string) + 1000 × 4 bytes (the indices) = ~5 KB
200× compression for the common shared-cause failure case. mongo doesn't deduplicate strings within a document; every byte matters at rest and on every fetch.

3. vol_index / resource_index are forward refs, not a wire optimization.

These aren't about saving bytes. They're how stork's reconciler answers "which entries does this VM own?" without rebuilding ownership from backup metadata on every reconcile pass. At 1000 VMs × 10000 resources, rebuilding the ownership graph on every reconcile would be expensive. Persisting the per-VM forward-refs lets the reconciler do an O(1) lookup.

Without these forward-refs, every cascade rule (Rule 1: volume failure → owner VMs; Rule 2: skip resources of failed VMs; Rule 3: resource failure → owner VMs; cleanup) needs to re-derive ownership. The indices are a write-time algorithmic data structure that the GetRestoreResourceDetails API doesn't replace.

4. Pattern consistency with PB-16170.

The resource-split work already established indexed reasons[] + reason_index on RestoreInfo.Resources. Keeping the same pattern here is consistent. Diverging would mean reasoning differently about identical compactness concerns.

5. Per-PR-comment trade-off if we drop indices:

If we resolve everything to strings on RestoreInfo:

Inspect response size grows 5-200× depending on duplicate-reason ratio
mongo storage grows the same
stork has to rebuild ownership graph on every reconcile (no forward-refs)
@vsundarraj-px @siva-portworx

pkg/apis/v1/api.proto
kgarg-px: addressed

pkg/apis/v1/api.proto
kgarg-px: Raised a JIRA for it : https://purestorage.atlassian.net/browse/PB-16708

pkg/apis/v1/api.proto
```diff
@@ -1366,6 +1419,18 @@ message RestoreInfo {
         bool skip_vm_restart = 2;
     }
 
+    // Per-VM persistent shape mirrored from stork CR.
+    // Returned by Restore.Inspect; matches mongo storage shape.
+    // reason_index is a 0-based position into RestoreInfo.reasons[].
+    // vol_index / resource_index are 0-based positions into Status.Volumes[] /
+    // Status.Resources[] (append-only — those arrays must never reorder).
+    message VirtualMachineStatus {
+        VirtualMachine    vm             = 1;
```
vsundarraj-px: please make these field [ (gogoproto.embed) = true ]; else it will be under nested VirtualmachineDetail.VirtualMachine.Name. with above it will be VirtualMachineDetail.Name

vsundarraj-px (approved):

vsundarraj-px (approved):
LGTM

pkg/apis/v1/api.proto
```diff
@@ -1379,6 +1452,15 @@ message RestoreInfo {
     // retention_period stores the object-lock retention period in days for the restore
     // This is only set when the backup location has object lock enabled
     int64 retention_period = 32;
+    // Total number of virtual machines in a virtual machine restore; 0 for
+    // namespace-scoped restores.
+    int64 virtual_machines_total_count = 35;
```
siva-portworx: Any reason why we are skipping 33 and 34 and moved on to 35?

pkg/apis/v1/api.proto
```diff
@@ -3380,6 +3480,83 @@ message BackupResourceObject {
...
+
+    // Aggregate per-VM counts for a virtual machine restore.
+    int64 virtual_machines_total_count  = 12;
+    int64 virtual_machines_failed_count = 13;
+    // Per-VM detail. Empty for namespace-scoped restores, or restores that do
+    // not provide per-VM detail.
+    repeated VirtualMachineDetail virtual_machines = 14;
+
+    // Filtered/paginated detail arrays.
+    repeated RestoreInfo.Volume volumes = 16;
```
siva-portworx: Same here, 15 is skipped.

pkg/apis/v1/api.proto
```diff
@@ -3380,6 +3480,83 @@ message BackupResourceObject {
...
+    // Restore uid
+    string uid = 3;
+    // Filter options
+    Filter filter = 4;
+    // Controls whether Successful entries are included in each per-VM
+    // VirtualMachineDetail.resources list. Default false → focused view
+    // (only Failed / Skipped / Retained / Deleted entries per VM). True →
+    // full view including derived Successful entries (for "View JSON" /
+    // detail-inspect use cases).
+    bool include_successful = 5;
```
siva-portworx: what will happen when some one in the cli give include_successful as false and give status filter as "success" Which will take priority? Can we document this behaviour to avoid confusion.

pkg/apis/v1/api.proto
```diff
@@ -1345,6 +1408,11 @@ message RestoreInfo {
             Success = 6;
             Retained = 7;
             PartialSuccess = 8;
+            Skipped = 9;
+            // Deleted indicates a successfully restored volume or resource
+            // entry that was later removed during cleanup because its owning
+            // object ultimately failed.
+            Deleted = 10;
```
siva-portworx, line 1417: Can Deleted and Skipped be restore top level status? Currently we are mixing status of resource and restore status. Should we mention in the comment to mention that it is applicable for restore object or resource ?

siva-portworx (commented):
LGTM to me after address the comments.

pkg/apis/v1/api.proto
kgarg-px: addressed

pkg/apis/v1/api.proto
kgarg-px: addressed

pkg/apis/v1/api.proto
kgarg-px: addressed, added comments

pkg/apis/v1/api.proto
kgarg-px: removed, not needed already present in filter message

kgarg-px:
Addressed all the review comments and concluded the design, merging the PR

PR #434, opened 2026-07-13
PB-15630: add Restore.GetRestoreCRDownloadLocation RPC + proto messages

lalat-das:
- New unary RPC returns access details for a restore CR JSON
- SignedURLLocation (URL + expiry) for S3/Azure/Google
- NFSLocation (server/export/path) for NFS.
- REST: GET /v1/restore/{org_id}/{name}/crDownloadLocation.

Closes # pb-15630

the test of the header changes will appear in px-backup PR.


vsundarraj-px (approved):
LGTM.
nit: The DownloadLocation() seems little miss leading. We are giving file path. But I feel bundling NFS inside it is making it this way. We give nfs path for entire location not just the path of the file. 

I would like to see some better name but other wise everything else looks fine.

pkg/apis/v1/api.proto
```diff
@@ -3610,6 +3610,109 @@ service Restore {
...
+    // Restore name.
+    string name = 2;
+
+    // Restore uid.
+    string uid = 3;
+
+    // Signed URL lifetime in seconds. Applies ONLY to S3 / Azure /
+    // Google backends. Server clamps to [60, 604800] (1 min to 7 days).
+    // Zero or unset -> default 1800 (30 minutes). Ignored for NFS.
+    uint64 expiry_seconds = 4;
```
siva-portworx: From auggie: 
```
   🟡 Medium: expiry_seconds is uint64 — should be uint32 or int64

   uint64 for a value clamped to [60, 604800] (max ~600K) is unnecessarily wide. More importantly, uint64 has poor cross-language interoperability:

     • JavaScript/JSON: Numbers above 2^53 lose precision. gRPC-web and REST/JSON clients will receive expiry_seconds as a JSON string (per proto3 JSON mapping for uint64), which is awkward for
       a simple integer field.
     • The swagger spec already shows "format": "uint64" with "type": "string" — clients will need to parse a string for what should be a small number.

   Consider uint32 (max ~4B, more than enough) or int32 (conventional for seconds-based durations). This avoids the JSON string encoding quirk.
```

siva-portworx (approved):

pkg/apis/v1/api.proto
lalat-das: make sense. Will change the access specifier.

lalat-das:
> LGTM. nit: The DownloadLocation() seems little miss leading. We are giving file path. But I feel bundling NFS inside it is making it this way. We give nfs path for entire location not just the path of the file.
> 
> I would like to see some better name but other wise everything else looks fine.

As you guessed because of NFS inclusion made it look like this. it is like I am extracting /downloading the location details. and eventually downloading the actual file from the location. essentially the resigned URL is also location detail of the file.

But I would love to name the way you feel like , please suggest a name, I will do the nam-karan as you wish..

PR #435, opened 2026-07-22
PB-16856: add Harvester provider to ClusterInfo Provider enum

sodey-px:
Adds a new Harvester value (= 6) to the ClusterInfo.Provider enum in api.proto, representing Harvester HCI clusters (which run on RKE2). 
This allows the backup API to identify and filter clusters whose cloud provider is Harvester, alongside the existing AWS, Azure, Google, IBM, and Rancher providers.

https://purestorage.atlassian.net/browse/PB-16856


pkg/apis/v1/api.proto
```diff
@@ -144,14 +144,15 @@ message ClusterInfo {
...
-    // Cloud provider type
+    // Cloud provider type. Harvester is an HCI cluster built on RKE2.
     enum Provider {
         Invalid = 0;
         AWS = 1;
         Azure = 2;
         Google = 3;
         IBM = 4;
         Rancher = 5;
+        Harvester = 6;
```
adasingh-px, line 156: [ignorable nit]: We can move the comment here: // Harvester is an HCI cluster built on RKE2. 
Instead of the heading.

adasingh-px (approved):

PR #437, opened 2026-07-28
added Proto changes needed for Capturing Volume dependencies in Mongo

apatani-px:
JIRA: https://purestorage.atlassian.net/browse/PB-16076

Design doc : https://purestorage.atlassian.net/wiki/spaces/PXBAC/pages/5262213284/Scale+Performance+Delete+Path+Optimization

To be reviewed along with : https://github.com/pure-px/px-backup/pull/2802


pkg/apis/v1/api.proto
```diff
@@ -1172,6 +1174,27 @@ message BackupInfo {
...
+        // dependent_snapshot_id is set while this volume is blocked by a newer
+        // incremental
+        string dependent_snapshot_id = 1;
+
+        // retry_count is bumped by the worker defer block on panic/error. Informational
+        // only
+        int32 retry_count = 2;
+
+        // error_reason captures the last failure message. Updated on every retry.
+        string error_reason = 3;
```
siva-portworx: Since this persisting of information is for internal use only, we can consider to have a indexed error reason to save space instead of repeated error string in all the volumes

pkg/apis/v1/api.proto
```diff
@@ -1172,6 +1174,27 @@ message BackupInfo {
...
+
+        // UID of backup owning the blocker snapshot
+        string dependent_backup_uid = 4;
+
+        // Name of backup owning the blocker snapshot (for indexed lookup)
+        string dependent_backup_name = 5;
```
siva-portworx: Same thing for the dependent backup name and uid. Most of the case, it will be same backup name and uid. let us discuss on this as well

pkg/apis/v1/api.proto
```diff
@@ -1172,6 +1174,27 @@ message BackupInfo {
...
 
+    // CleanupDetails captures per-volume garbage-collection state for the unified
+    // deletion pipeline. All fields are optional.
+    message CleanupDetails {
+        // dependent_snapshot_id is set while this volume is blocked by a newer
+        // incremental
+        string dependent_snapshot_id = 1;
+
+        // retry_count is bumped by the worker defer block on panic/error. Informational
+        // only
```
siva-portworx: The comment is not clear on the usecase of the variable. Can you please add some details on when the retry count will be updated

apatani-px: removed it since it is of no use

apatani-px:
raised new one , dont review

PR #439, opened 2026-07-28
PB-16527 enable manual sync in non-federated env

mrajendra-px:
# Summary
https://purestorage.atlassian.net/browse/PB-16527

# Testing
https://github.com/pure-px/px-backup/pull/2796


pkg/apis/v1/api.proto
```diff
@@ -799,15 +799,25 @@ message BackupLocationInfo {
...
+    // Populated for both Automatic and Manual BLs in both Federated and Non-Federated modes.
     SyncInfo sync_info = 14;
+    // sync_manual controls whether this BackupLocation uses manual or automatic sync.
+    // - true  : Manual mode. Periodic BackupSync is skipped for this BL.
+    //           Syncs only occur when explicitly triggered via sync=true.
+    // - false : Automatic mode. Periodic BackupSync runs every 10 minutes.
+    // - null (not set): rejected on Create; treated as no-op on Update.
+    // Valid in both Federated and Non-Federated deployment modes.
+    // Uses BoolValue wrapper to distinguish between "not provided" (nil) and "explicitly set to false".
+    google.protobuf.BoolValue sync_manual = 16;
```
siva-portworx, line 820: Any reason why we are skipping index 15?

pkg/apis/v1/api.proto
```diff
@@ -799,15 +799,25 @@ message BackupLocationInfo {
...
     // sync_info provides information about the sync operation including status, timing, and stats.
+    // Populated for both Automatic and Manual BLs in both Federated and Non-Federated modes.
     SyncInfo sync_info = 14;
+    // sync_manual controls whether this BackupLocation uses manual or automatic sync.
+    // - true  : Manual mode. Periodic BackupSync is skipped for this BL.
+    //           Syncs only occur when explicitly triggered via sync=true.
+    // - false : Automatic mode. Periodic BackupSync runs every 10 minutes.
+    // - null (not set): rejected on Create; treated as no-op on Update.
+    // Valid in both Federated and Non-Federated deployment modes.
+    // Uses BoolValue wrapper to distinguish between "not provided" (nil) and "explicitly set to false".
```
siva-portworx, line 819: For old existing Backuplocation object in Federated and Non-Federated mode, what will be the value of this parameter? 
For federated mode, it will be migrated to true and for none federated mode it will be false.
Also for federated mode, the default value will be true, if the customer do not set it ( I means nil) And also we need to mention that in federated mode only manual is supported, automatic is not supported.

pkg/apis/v1/api.proto
```diff
@@ -833,6 +843,22 @@ message BackupLocationInfo {
...
+        // Total number of backup objects enumerated during the most recent
+        // sync cycle. Includes new imports, already-synced skips, and
+        // per-object errors. Updated periodically during InProgress (every
+        // 100 objects or 5 seconds, whichever first) so consumers can
+        // distinguish "sync progressing but nothing new" (processed > 0,
+        // synced == 0) from "sync stuck" (neither incrementing while
+        // status = InProgress).
+        // Zeroed at the start of each new sync cycle. Non-federated only.
+        // Limitation: only the most recent cycle is retained.
+        int64 backups_processed = 6;
```
siva-portworx, line 861: @mkoppal-px , @mrajendra-px both the backups_synced and backups_processed stats are useful for the identifying the issue with backupsync progress, even if it is triggered manual in the federated mode. If the time is the constraint because of which we are not doing it for ferdation mode, can we evaluate the effort for supporting it in federated mode and see whether it can be supported? Having this as zero for federated mode might be confusing to SAP customer.

pkg/apis/v1/api.proto
mrajendra-px: Tag 15 is already used by validation_started_at (line 801)

pkg/apis/v1/api.proto
mrajendra-px: Migration skips federated, so old BLs in federated remain nil.. during Create BL, in federated mode no check is present --- accepted & stored as nil

pkg/apis/v1/api.proto
mrajendra-px: these are all the changes i need to verify in federated mode (the first 3 are easy & will just take a day):

1. test migration of sync-manual to true in all old existing BLs
2. ensure the create continues to treat sync-manual as a non-mandatory parameter in federated (and we internally fill it as true)
3. ensure update and create does not allow sync-manual to False
4. The Stork BLs CR status currently only carries {Status, Reason, LastUpdateTime} — no counters are computed or exposed by Stork today. Maybe need to explore what Das was mentioning that day.... ~2 days to verify the design/process and test

the #4 is what i need some clarity on + timetaking

siva-portworx (approved):

PR #440, opened 2026-07-28
PB-16802: Use index and internal chunks instead of flat

kgarg-px:
Below changes to include just indexes and no flat resources on the response 
Closes #


pkg/apis/v1/api.proto
```diff
@@ -1404,6 +1398,12 @@ message RestoreInfo {
         repeated string zones = 7;
         map<string, string> options = 8;
         uint64 total_size = 9;
+        // Index into RestoreObject.resource_names for the PVC name.
```
vsundarraj-px: we dont need to change the current volume structure. pvc is already in the structure why do we need pvc_name_idx? source_namespace is already a field

pkg/apis/v1/api.proto
```diff
@@ -1446,10 +1446,23 @@ message RestoreInfo {
...
+    // Per-VM restore status. Identity fields (name, namespace) are inline;
+    // forward-refs (volume_index into RestoreInfo.volumes, resource_refs into
+    // the pooled resource name/type/namespace tables) locate owned entries
+    // without duplicating them on this message. Reason on the status
+    // sub-message is server-resolved from RestoreObject.reasons when
+    // constructing the wire response.
     message VirtualMachineStatusInfo {
-        VirtualMachine vm     = 1 [ (gogoproto.embed) = true ];
-        StatusInfo     status = 2;
+        StatusInfo status = 1;
```
vsundarraj-px: nit: move status below Name  just for readability purpose

pkg/apis/v1/api.proto
```diff
@@ -1446,10 +1446,23 @@ message RestoreInfo {
...
+        string name       = 2;
+        string namespace  = 3;
+        // Positions into RestoreInfo.volumes for volumes owned by this VM.
+        repeated int32 volume_index = 4;
+        // References into the pooled non-volume resources owned by this VM.
+        repeated ResourceRef resource_refs = 5;
+        // Wall-clock stamp taken when the VM entered InProgress; used by the
+        // worker-pool timeout enforcer.
+        google.protobuf.Timestamp entered_in_progress_at = 6;
```
vsundarraj-px: This just adds CR size or mongo load. I dont see a great benefit of this data.

pkg/apis/v1/api.proto
```diff
@@ -1541,6 +1554,8 @@ message RestoreInfo {
                 string name = 1;
                 // Index into the reasons table in RestoreResources (0 means no reason)
                 int32 reason_index = 2;
+                // Index into RestoreObject.resource_names for the resource name.
+                int32 name_idx = 3;
```
vsundarraj-px: name is already part of the same field at line 1554 why index for it?

pkg/apis/v1/api.proto
```diff
@@ -1599,9 +1614,15 @@ message RestoreObject {
     Metadata metadata = 1 [ (gogoproto.embed) = true ];
     RestoreInfo restore_info = 2 [ (gogoproto.embed) = true ];
     // For internal use only; not returned in API responses.
-    repeated VirtualMachineStatusIndexInfo vm_status_index = 3;
+    repeated string reasons = 3;
+    // For internal use only; not returned in API responses.
+    repeated string namespaces = 4;
+    // For internal use only; not returned in API responses.
+    repeated string resource_types = 5;
```
vsundarraj-px: These are not required as well. resource_types resource_names drivers

pkg/apis/v1/api.proto
kgarg-px: addressed

pkg/apis/v1/api.proto
kgarg-px: addressed

pkg/apis/v1/api.proto
kgarg-px: addressed

pkg/apis/v1/api.proto
kgarg-px: addressed

vsundarraj-px (approved):
LGTM for Vm restore. RestoreResourceObject should include FileLevelRestoreStatusInfo from RestoreInfo. Please open a new task so that this is added.

PR #441, opened 2026-07-30
PB-17039: Add filters for BackupDeleteRequest

shhegde-px:
Adds the 7 UI-required filters (status, backup_object_type, labels, time_range, owners, schedule_policy_ref, backup_schedule_ref) to BackupDeleteRequest so bulk backup deletion can be narrowed the same way Backup.Enumerate already is, and regenerates the corresponding Go/swagger code.

Filters are in a dedicated options message (following the ClusterEnumerateOptions / BackupLocationEnumerateOptions convention), and backup_object_type / statuses are enum-based, reusing the existing BackupInfo.BackupObjectType and BackupInfo.StatusInfo.Status rather than plain strings. No migration is needed since these live on a request message and the persisted BackupInfo fields already use the same enums.


sodey-px (approved):

pkg/apis/v1/api.proto
```diff
@@ -3270,6 +3270,26 @@ message BackupDeleteRequest {
...
+    // filter to use backupObjectType on object
+    string backup_object_type = 17;
+    // filter based on the object status.
+    // Need to pass the string value of the status field of the object ( backup /restore )
+    // For example, {"Success", "Failed"}
+    repeated string status = 18;
+    // filter to use policy name and uid. Any object that contains the filter will be returned.
+    repeated ObjectRef schedule_policy_ref = 19;
+    // filter to use backup schedule policy name and uid. Any object that contains the filter will be returned.
+    repeated ObjectRef backup_schedule_ref = 20;
```
adasingh-px: Create another custom field for all these new filters: `BackupDeleteEnumerateOptions` similar to how it is being followed for: `VolumeResourceOnlyPolicyEnumerateOptions` and `ClusterEnumerateOptions`

adasingh-px (approved):

pkg/apis/v1/api.proto
```diff
@@ -3232,6 +3232,29 @@ message BackupInspectRequest {
...
+message BackupDeleteEnumerateOptions {
+    // label selectors for the object for filtering
+    map<string, string> labels = 1;
+    // Time range for which to delete objects
+    TimeRange time_range = 2;
+    // Filter to use owner uids on objects. Any object that contains the
+    // filter will be returned
+    repeated string owners = 3;
+    // filter to use backupObjectType on object
+    string backup_object_type = 4;
```
adasingh-px: What details we accept when in backup_object_type?
Should we move to enum based only?

I see there is a enum available: 
```
    message BackupObjectType {
        Type type = 1;
         enum Type {
             Invalid = 0;
             All = 1; // for All application backup
             VirtualMachine = 2; // virtualMachine for VirtualMachine specific backup
         }
     }
```

Can you please see, if we can keep it common, and see no migration is needed. Hopefully a few changes in px-backup would work? Just see if the same is not present in MongoDB.

CC: @sodey-px , @vikasit12 , @mkoppal-px

adasingh-px (changes requested):

pkg/apis/v1/api.proto
```diff
@@ -3232,6 +3232,29 @@ message BackupInspectRequest {
...
+    // filter based on the object status.
+    // Need to pass the string value of the status field of the object ( backup /restore )
+    // For example, {"Success", "Failed"}
+    repeated string status = 5;
```
vikasit12: if repeated string owners then it should be repeated string statuses.
Also status is an enum too and narrow down the possibility of error we should make in enum

pkg/apis/v1/api.proto
vikasit12: You can refer https://github.com/portworx/px-backup-api/commit/dad0f2ab3caf8fb5d2fe0d5666292728ea42ecda where similar change has been made

vikasit12:
@shhegde-px Please resolve the merge conflict

vikasit12 (approved):

adasingh-px (approved):

PR #443, opened 2026-08-05
PB-16802: Add size of per pvc during restore for

kgarg-px:
getRestoreResourceDetails API reqs

In the UI we need to show per pvc size as well under getRestoreResourceDetails API response.
current is below:
image

Closes #


kgarg-px:
working as expected
image
image

aks-px (approved):

